PR #207, opened 2021-12-03
chore: Adjust nuget override

jeromelaban:


nickrandolph (approved):

src/Directory.Build.props
```diff
@@ -102,7 +102,4 @@
 		<DefaultItemExcludes>$(DefaultItemExcludes);obj/**;bin/**</DefaultItemExcludes>
 	</PropertyGroup>
 
-	<!-- <PropertyGroup Condition="'$(Configuration)'=='Debug'">
```
nickrandolph, line 105: @jeromelaban  The crosstargeting_override.props.sample doesn't include setting the UnoTargetFrameworkOverride value, should this be added to make it easier for devs to choose a specific TFM to build for?

Issue #374, opened 2022-04-25
Add IDispatcher


nickrandolph:
There's a need for an abstraction to allow viewmodel to invoke code on the UI thread.

## What would you like to be added:

IDispatcher - Uno.Extensions.Core (so it's accessible to all extensions)
Dispatcher - Uno.Extensions.Navigation.UI (as it's dependent on Window)

## Why is this needed:

Currently there's no way for viewmodel to execute code on the UI thread. This is required when using reactive commands where the code is executed on a background thread to avoid blocking

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renders for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks

PR #375, opened 2022-04-25
feat: Adding IDispatcher

nickrandolph:
GitHub Issue (If applicable): #374

## PR Type

- Feature

No way for viewmodel to execute code on UI thread

## What is the new behavior?

Viewmodel can take dependency on IDispatcher and use the ExecuteAsync method to invoke code on UI thread

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Extensions.Core/DispatcherExtensions.cs
+++ b/src/Uno.Extensions.Core/DispatcherExtensions.cs
@@ -1,0 +1,22 @@
+﻿namespace Uno.Extensions;
+
+public static class DispatcherExtensions
+{
+	public static Task ExecuteAsync(this IDispatcher dispatcher, Func<Task> action)
+	{
+		return dispatcher.ExecuteAsync(async () =>
+		{
+			await action();
+			return true;
+		});
+	}
+
+	public static Task ExecuteAsync(this IDispatcher dispatcher, Action action)
+	{
+		return dispatcher.ExecuteAsync(async () =>
+		{
+			action();
+			return true;
+		});
+	}
+}
--- a/src/Uno.Extensions.Core/DispatcherExtensions.cs
+++ b/src/Uno.Extensions.Core/DispatcherExtensions.cs
@@ -2,25 +2,25 @@
 
 public static class DispatcherExtensions
 {
-	public static Task ExecuteAsync(this IDispatcher dispatcher, Func<CancellationToken, Task> action, CancellationToken token)
+	public static async ValueTask ExecuteAsync(this IDispatcher dispatcher, AsyncAction action, CancellationToken token)
 	{
-		return dispatcher.ExecuteAsync(async (CancellationToken t) =>
+		await dispatcher.ExecuteAsync(async (CancellationToken t) =>
 		{
 			await action(t);
 			return true;
 		}, token);
 	}
 
-	public static Task ExecuteAsync(this IDispatcher dispatcher, Func<Task> action)
+	public static async ValueTask ExecuteAsync(this IDispatcher dispatcher, Func<ValueTask> action)
 	{
-		return dispatcher.ExecuteAsync(async (CancellationToken t) =>
+		await dispatcher.ExecuteAsync(async (CancellationToken t) =>
 		{
 			await action();
 			return true;
 		}, CancellationToken.None);
 	}
 
-	public static Task<TResult> ExecuteAsync<TResult>(this IDispatcher dispatcher, Func<Task<TResult>> actionWithResult)
+	public static ValueTask<TResult> ExecuteAsync<TResult>(this IDispatcher dispatcher, Func<ValueTask<TResult>> actionWithResult)
 	{
 		return dispatcher.ExecuteAsync(async (CancellationToken t) =>
 		{
@@ -28,18 +28,18 @@ public static class DispatcherExtensions
 		}, CancellationToken.None);
 	}
 
-	public static Task ExecuteAsync(this IDispatcher dispatcher, Action action)
+	public static async ValueTask ExecuteAsync(this IDispatcher dispatcher, Action action)
 	{
-		return dispatcher.ExecuteAsync(async (CancellationToken t) =>
+		await dispatcher.ExecuteAsync(async (CancellationToken t) =>
 		{
 			action();
 			return true;
 		}, CancellationToken.None);
 	}
 
-	public static Task ExecuteAsync(this IDispatcher dispatcher, Action<CancellationToken> action, CancellationToken token)
+	public static async ValueTask ExecuteAsync(this IDispatcher dispatcher, Action<CancellationToken> action, CancellationToken token)
 	{
-		return dispatcher.ExecuteAsync(async (CancellationToken t) =>
+		await dispatcher.ExecuteAsync(async (CancellationToken t) =>
 		{
 			action(t);
 			return true;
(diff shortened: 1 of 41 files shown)


src/Uno.Extensions.Core/IDispatcher.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Uno.Extensions;
+
+public interface IDispatcher
+{
+	Task<TResult> ExecuteAsync<TResult>(Func<Task<TResult>> actionWithResult);
```
jeromelaban, line 5: You may want to provide a CancellationToken as well.

src/Uno.Extensions.Navigation.UI/CoreDispatcherExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+	public static async Task<TResult> ExecuteAsync<TResult>(this Windows.UI.Core.CoreDispatcher dispatcher, Func<Task<TResult>> actionWithResult)
+	{
+		var completion = new TaskCompletionSource<TResult>();
+		await dispatcher.RunAsync(Windows.UI.Core.CoreDispatcherPriority.Normal, async () =>
+		{
+			try
+			{
+				var result = await actionWithResult();
+				completion.SetResult(result);
```
jeromelaban, line 14:
```suggestion
				completion.TrySetResult(result);
```

src/Uno.Extensions.Navigation.UI/CoreDispatcherExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+			}
+			catch (Exception ex)
+			{
+				completion.SetException(ex);
```
jeromelaban, line 18:
```suggestion
				completion.TrySetException(ex);
```

cc @dr1rrb

src/Uno.Extensions.Core/IDispatcher.cs
nickrandolph: Yep, good idea

nickrandolph: I'll go ahead and add it today

src/Uno.Extensions.Core/DispatcherExtensions.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace Uno.Extensions;
+
+public static class DispatcherExtensions
+{
+	public static Task ExecuteAsync(this IDispatcher dispatcher, Func<CancellationToken, Task> action, CancellationToken token)
```
francoistanguay, line 5: I think we have some standard delefates like FuncAsync or AsyncFunc for the CT/Task signature

src/Uno.Extensions.Core/IInstanceRepository.cs
```diff
@@ -3,7 +3,20 @@
 
 namespace Uno.Extensions;
 
-public interface IInstanceRepository
+internal interface IInstanceRepository
 {
     IDictionary<Type, object> Instances { get; }
 }
+
+internal interface ITransientInstanceRepository : IInstanceRepository
```
francoistanguay, line 11: What is this for? Let's discuss

francoistanguay (approved):

src/Uno.Extensions.Core/DispatcherExtensions.cs
dr1rrb: https://github.com/unoplatform/uno.extensions/blob/main/src/Uno.Extensions.Core/AsyncAction.cs :)

src/Uno.Extensions.Core/IDispatcher.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Uno.Extensions;
+
+public interface IDispatcher
```
dr1rrb, line 3: For perf consideration, I would add `void Enqueue(Action action, CancellationToken ct)` on the interface. This will avoid to create a `TaskCompletionSource` when not needed. 

Maybe we should also add a `void Enqueue(ActionAsync action, CancellationToken ct)` which would add a `try/catch` around the action an log exception to prevent silent/crashy exceptions in `async void`.

Both could be flagged with the `[EditorBrowsable(EditorBrowsableState.Advanced)]` attribute.

src/Uno.Extensions.Navigation.UI/CoreDispatcherExtensions.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Uno.Extensions;
+
+#if !WINUI
+public static class CoreDispatcherExtensions
```
dr1rrb, line 4:
```suggestion
internal static class CoreDispatcherExtensions
```

src/Uno.Extensions.Navigation.UI/ServiceProviderExtensions.cs
```diff
@@ -11,43 +11,90 @@ public static void AttachServices(this Window window, IServiceProvider services)
 
 	internal static IServiceProvider RegisterWindow(this IServiceProvider services, Window window)
 	{
-		return services.AddInstance(window);
+		var sp = services.AddScopedInstance(window);
+		// Force a get on the dispatcher to make sure the instance is created on this
+		// thread, which should ensure the DispatcherQueue exists
```
dr1rrb, line 16: This assume the current implementation of the `Dispatcher` class (which do a `GetForCurrentThread()` when it can't resolve it from the `Window`). I would prefer to explicitly do something like `services.AddScopedInstance(new Dispatcher(window.Dispatcher));`. 

Also then, could we remove the `Window` from the `services`?

dr1rrb (commented):
Could please also add XML doc to new public APIs ?

src/Uno.Extensions.Core/DispatcherExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+	{
+		return dispatcher.ExecuteAsync(async (CancellationToken t) =>
+		{
+			await action(t);
+			return true;
+		}, token);
+	}
+
+	public static Task ExecuteAsync(this IDispatcher dispatcher, Func<Task> action)
```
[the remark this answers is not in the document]
nickrandolph: I agree from a best practice perspective we should be encouraging the use of cancellationtokens for async actions but the reality is that if devs have to specify a cancellation token (and provide an action that accepts a cancellation token) that introduces a lot of code clutter.

src/Uno.Extensions.Core/DispatcherExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+	public static Task ExecuteAsync(this IDispatcher dispatcher, Action action)
+	{
+		return dispatcher.ExecuteAsync(async (CancellationToken t) =>
+		{
+			action();
+			return true;
+		}, CancellationToken.None);
+	}
+
+	public static Task ExecuteAsync(this IDispatcher dispatcher, Action<CancellationToken> action, CancellationToken token)
```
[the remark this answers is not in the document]
nickrandolph: The scenario here is that you just have a bunch of synchronous code that you want to execute on the UI thread. Unless I'm missing something we're not creating a Task that isn't being awaited

src/Uno.Extensions.Core/IDispatcher.cs
nickrandolph: I've pushed these suggestions to a new issue #385

PR #387, opened 2022-04-27
fix(feeds): Miscs fixes for feeds

dr1rrb:
Miscs fixes for feeds

### DIFF
--- a/src/Uno.Extensions.Reactive.UI/Presentation/Bindings/Collections/NullObservableVector.cs
+++ b/src/Uno.Extensions.Reactive.UI/Presentation/Bindings/Collections/NullObservableVector.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Windows.Foundation.Collections;
+
+namespace Uno.Extensions.Reactive.Bindings.Collections
+{
+	/// <summary>
+	/// Null pattern implementation of <see cref="IObservableVector{T}"/>.
+	/// </summary>
+	internal class NullObservableVector<T> : IObservableVector<T>
+	{
+#pragma warning disable CS0067 // Event not raised
+		/// <inhertidoc />
+		public event VectorChangedEventHandler<T>? VectorChanged;
+#pragma warning restore CS0067
+
+		/// <inhertidoc />
+		public int Count { get; } = 0;
+
+		/// <inhertidoc />
+		public bool IsReadOnly { get; } = true;
+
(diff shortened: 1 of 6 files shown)


src/Uno.Extensions.Reactive/Core/Axes/MessageAxis.T.cs
```diff
@@ -35,7 +35,7 @@ internal MessageAxis(string name, Aggregator aggregate)
 	/// <param name="value">The metadata to encapsulate..</param>
 	/// <returns>The raw axis value.</returns>
 	[Pure]
-	public MessageAxisValue ToMessageValue(T? value)
+	public MessageAxisValue ToMessageValue(T value)
```
nickrandolph, line 38: Why isn't T marked as nullable?

nickrandolph, line 38: The naming for the From and To methods might be better as ToAxisValue and FromAxisValue

src/Uno.Extensions.Reactive.UI/Presentation/Bindings/Collections/NullObservableVector.cs
```diff
@@ -0,0 +1,75 @@
...
+		IEnumerator IEnumerable.GetEnumerator() => new Enumerator();
+
+		/// <inhertidoc />
+		public void Add(T item) => throw new NotSupportedException();
+		
+		/// <inhertidoc />
+		public void Clear() => throw new NotSupportedException();
+
+		/// <inhertidoc />
+		public bool Contains(T item) => false;
```
nickrandolph, line 45: Is there a reason that Clear throws an exception whereas Contains returns false? Shouldn't Contains throw a NotSupportedException too? Same for CopyTo, Remove etc

nickrandolph:
@dr1rrb do we need to get these fixed merged as part of the release this week?

src/Uno.Extensions.Reactive.UI/Presentation/Bindings/Collections/NullObservableVector.cs
dr1rrb: This collection is flagged as `ReadOnly` so we throw exception in all methods that attempt to modify the collection (i.e. write), while we do allow "read" methods.

nickrandolph (approved):

dr1rrb:
Yes it is, but only impacts grouped collection which are not publicly exposed yet ... needs to rebase and merge

PR #388, opened 2022-04-28
feat(feed): Add ability to data-bind to feed and state exposed by VM

dr1rrb:
Add ability to data-bind to feed and state exposed by VM

(i.e. deprecating the IInput pattern)


carldebilly (approved):

src/Uno.Extensions.Reactive.UI/Presentation/Bindings/Bindable.cs
```diff
@@ -12,18 +13,20 @@ namespace Uno.Extensions.Reactive.Bindings;
...
 /// </summary>
 /// <typeparam name="T">The type of the value</typeparam>
-public class Bindable<T> : IBindable, INotifyPropertyChanged
+public class Bindable<T> : IBindable, INotifyPropertyChanged, IFeed<T>
 {
 	/// <inheritdoc />
 	public event PropertyChangedEventHandler? PropertyChanged;
 
-	private T? _value;
+	private T _value = default!; // This is going to be init by property.Subscribe(OnOwnerUpdated);, and anyway with bindings we cannot ensure non-null!
```
jeromelaban, line 21: If it can't be non-null, why keep it non-nullable?

dr1rrb: Otherwise I'm not able to re-expose it as `IFeed<T>` (would be `IFeed<T?>`)

Issue #473, opened 2022-05-20
[Feed] Generate mapping of VM => BindableVM for nav


dr1rrb:
[Feed] Generate mapping of VM => BindableVM for nav

PR #522, opened 2022-06-03
feat(feed): Generate the BinVM mapping for nav

dr1rrb:
fixes https://github.com/unoplatform/uno.extensions/issues/473

## Feature
Generate a map from VM to it's bindable VM type

Users have to declare there the mapping in there app (or deal with BindableVM)

## What is the new behavior?
Mapping is generated and nav will use it to avoid devs to bother of the generated VMs

## PR Checklist


src/Uno.Extensions.Reactive.Generator/Bindables/BindableViewModelGenerator.cs
```diff
@@ -32,13 +34,15 @@ where IsSupported(type)
 
 		foreach (var vm in viewModels)
 		{
-			yield return (vm, Generate(vm));
+			yield return (vm.ToString(), Generate(vm));
```
nickrandolph, line 37: Should this be the Name property? I think it'd be better to be explicit about which property is being used here, rather than using ToString and assuming it's returning a specific property value

nickrandolph (approved):

src/Uno.Extensions.Reactive.Generator/Bindables/BindableViewModelMappingGenerator.cs
```diff
@@ -0,0 +1,47 @@
...
+
+	public void Register(INamedTypeSymbol viewModelType, string bindableViewModelType)
+		=> _bindableVMs[viewModelType] = bindableViewModelType;
+
+	public (string file, string code) Generate()
+	{
+		var code = $@"#nullable enable
+			#pragma warning disable
+
+			namespace {_ctx.Context.Compilation.Assembly.Name}
```
nickrandolph, line 28: Is there no way to pick up the default namespace for the assembly? I'd prefer not to rely on the name of the assembly to imply namespace

dr1rrb: I would prefer also, but the default namespace is actually a VS property, so there is no reliable way to get (especially when building using CLI instead of VS), so I prefer to used the Assembly name. Anyway I think that the vast majority of project does not use it.

src/Uno.Extensions.Reactive.Generator/Bindables/BindableViewModelGenerator.cs
dr1rrb: I usually agree, but for the `INamedTypeSymblol` the `ToString()` is documented to return the fullname in a readable way (i.e. `<>` instead of ` ``1 `) and it's the common/right way to deal with it in code gen. The `Name` is only the class name, which could conflict with classes in different namespaces.

src/Uno.Extensions.Reactive.Generator/Bindables/BindableViewModelMappingGenerator.cs
nickrandolph: The issue with this is that it requires all head projects to output the same assembly name (in the case that your viewmodels are in the shared/ui project). Not a major issue but imho not a good reason to force all head projects to use the same assembly name

Issue #659, opened 2022-08-16
[Configuration] ImmutableList parameter can't be modified in IWritableOptions's UpdateAsync  


thierry-fortin1:
## Context
I have a IWritableOptions<HosOptions>: 
![image](https://user-images.githubusercontent.com/105242267/184912536-f90e22a3-8a36-4de6-92bc-d6a98c5e3ab2.png)
And HosOptions is that: 
![image](https://user-images.githubusercontent.com/105242267/184912721-05da9fef-ae82-4b1b-912c-67b9151495d0.png)
I don't have any problems with IWritableOptions when the parameters aren't lists.

## Current behavior
I want to update _hosConfig.Value.HiddenZones (add or remove a item in the ImmutableList, depending on if HiddenZones already has the zone in its list).

ImmutableList<OperatingZone> (doesn't work)
![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/105242267/184912343-fa75cb16-b52f-44ae-be95-35448aa0bd14.png)
![image](https://user-images.githubusercontent.com/105242267/184905642-909cbf92-d08f-4da9-b604-be64607d8412.png)

List<OperatingZone> (works)
![image](https://user-images.githubusercontent.com/105242267/184917392-3c281fb7-c47e-4799-bb46-2cc980b3fe11.png)
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/105242267/184911906-a0db3bdb-1ba4-487b-ad4f-21e0ffe9a941.png)
![image](https://user-images.githubusercontent.com/105242267/184916570-aaabc2a2-a5dc-44ff-85a1-8f720a121cb7.png)

## Expected behavior
Updating a ImmutableList should work in UpdateAsync, in this example, HiddenZones should be modified.

## Environment

Nuget Package:
2.2.0-dev.79

Package Version(s):

Affected platform(s):

- [X] iOS
- [X] Android
- [ ] WebAssembly
- [ ] WebAssembly renders for Xamarin.Forms
- [X] Windows
- [ ] Build tasks

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [X] 2022 (version: 17.3.0)
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

thierry-fortin1:
@dr1rrb

dr1rrb:
[...]

PR #662, opened 2022-08-17
feat: Handling serialization of immutablelist in configuration/settings

nickrandolph:
GitHub Issue (If applicable): #659

## PR Type

- Feature

UpdateAsync has a non-immutable option for updating settings
ImmutableList is not supported

## What is the new behavior?

The non-immutable overload has been removed
Immutabelist can be used in settings/configuration entity

## PR Checklist

## Other information


src/Uno.Extensions.Navigation.UI/Navigators/TypeExtensions.cs
```diff
@@ -6,12 +6,11 @@ public static class TypeExtensions
 {
 	public static ConstructorInfo? GetNavigationConstructor(this Type type, INavigator navigator, IServiceProvider services, out object[] constructorArguments)
 	{
-		var ctrs = type.GetConstructors();
-		foreach (var ctr in ctrs)
+		var ctr = type.GetConstructors().FirstOrDefault();
```
dr1rrb, line 9: Should we be able to configure which ctor to use?

src/Uno.Extensions.Configuration/ConfigurationBinder.cs
```diff
@@ -0,0 +1,731 @@
...
+	}
+
+	private static object BindImmutableList(
+		object collection,
+			Type collectionType,
+		IConfiguration config, BinderOptions options)
+	{
+		// ICollection<T> is guaranteed to have exactly one parameter
+		Type itemType = collectionType.GenericTypeArguments[0];
+		MethodInfo addMethod = collectionType.GetMethod("Add", DeclaredOnlyLookup);
```
dr1rrb, line 518: Should be great to use the `Builder` when available (NTH - the real fix would be to use the `System.Text.Json`).

src/Uno.Extensions.Configuration/ConfigurationBinder.cs
```diff
@@ -0,0 +1,731 @@
+﻿
+
+using System.Collections.Immutable;
+using System.ComponentModel;
+
+namespace Uno.Extensions.Configuration.Internal;
+
+public static class ConfigurationBinder
```
dr1rrb, line 8: Indeed LOT of reflection here ... would be great to replace that by `System.Text.Json` (in an other PR)

dr1rrb (approved):

src/Uno.Extensions.Navigation.UI/Navigators/TypeExtensions.cs
nickrandolph: I can't see a strong use case for supporting multiple constructors for view model creation. In the case where different data can be passed into the same view model type, you could have multiple optional parameters on the constructor

src/Uno.Extensions.Configuration/ConfigurationBinder.cs
nickrandolph: Yeh, there's a trade off between converting configuration section to a string and then parsing using System.Text.Json versus the current logic that uses reflection. Loading configuration isn't done that frequently, and the conversion to strongly typed entity (ie this binder logic) isn't done immediately on startup, only when the entities are actually requested

Issue #695, opened 2022-09-02
[Configuration] Getting IO  exception - Sharing violation when reading/writing configuration settings


nickrandolph:
## Current behavior

Use WritableOptions to read/write settings from multiple threads concurrently will cause IOException

## Expected behavior

Read/Write should not throw exception

PR #696, opened 2022-09-02
fix: Preventing read/write lock on writable settings

nickrandolph:
GitHub Issue (If applicable): #695

## PR Type

- Bugfix

There's no locking around file access for either read or write

## What is the new behavior?

Added semaphore around reading and writing to prevent concurrent access to same files

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Extensions.Configuration/Reloader.cs
+++ b/src/Uno.Extensions.Configuration/Reloader.cs
@@ -61,7 +61,7 @@ public class Reloader
 						}
 						if (!read)
 						{
-							await Task.Yield();
+							await Task.Delay(50);
 						}
 					}
 				}
--- a/src/Uno.Extensions.Configuration/Reloader.cs
+++ b/src/Uno.Extensions.Configuration/Reloader.cs
@@ -2,6 +2,9 @@
 
 public class Reloader
 {
+	internal static SemaphoreSlim ReadWriteLock = new SemaphoreSlim(1);
+	private const int MaxReadRetries = 100;
+
 	private ILogger Logger { get; }
 
 	private IConfigurationRoot Config { get; }
@@ -32,10 +35,35 @@ public class Reloader
 				}
 				else
 				{
-					var contents = File.ReadAllText(info.PhysicalPath);
-					if (Logger.IsEnabled(LogLevel.Debug)) Logger.LogDebugMessage($@"Contents '{contents}'");
-					if (Logger.IsEnabled(LogLevel.Debug)) Logger.LogDebugMessage($@"Loading from full path '{info.PhysicalPath}'");
-					fp.Load();
+					if (Logger.IsEnabled(LogLevel.Debug))
+					{
+						var contents = File.ReadAllText(info.PhysicalPath);
+						Logger.LogDebugMessage($@"Contents '{contents}'");
+						Logger.LogDebugMessage($@"Loading from full path '{info.PhysicalPath}'");
+					}
+					var read = false;
+					var attempt = 0;
+					while (!read && attempt++ < MaxReadRetries)
+					{
+						await ReadWriteLock.WaitAsync();
+						try
+						{
+							fp.Load();
+						}
+						catch (IOException)
+						{
+							// Only retry on IOExceptions eg sharing violation
+							read = false;
+						}
+						finally
+						{
+							ReadWriteLock.Release();
+						}
+						if (!read)
+						{
+							await Task.Yield();
+						}
+					}
 				}
 			}
 		}
--- a/src/Uno.Extensions.Configuration/Reloader.cs
+++ b/src/Uno.Extensions.Configuration/Reloader.cs
@@ -3,7 +3,6 @@
 public class Reloader
 {
 	internal static SemaphoreSlim ReadWriteLock = new SemaphoreSlim(1);
-	private const int MaxReadRetries = 100;
 
 	private ILogger Logger { get; }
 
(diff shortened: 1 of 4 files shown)


src/Uno.Extensions.Configuration/Reloader.cs
```diff
@@ -32,10 +35,35 @@ public async Task ReloadAllFileConfigurationProviders(string? configFile = defau
...
+						}
+						finally
+						{
+							ReadWriteLock.Release();
+						}
+						if (!read)
+						{
+							await Task.Yield();
+						}
+					}
```
jeromelaban, line 66: This may cause a live lock (two or more readers interleaving) with high CPU usage as a result. You may want to use an actual `Task.Delay(50)` instead to introduce entropy.

src/Uno.Extensions.Configuration/WritableOptions.cs
```diff
@@ -61,7 +63,31 @@ public async Task UpdateAsync(Func<T, T> applyChanges)
...
+			}
+			finally
+			{
+				Reloader.ReadWriteLock.Release();
+			}
+			if(!written)
+			{
+				await Task.Delay(50);
+			}
+		}
```
carldebilly, line 89: The code need to re-read the content of the file and always reapply the update on the newest version of the file instead of trying to always rewrite the same value, which is potentially obsoleted.

nickrandolph: I think give we have locks around the write operation, that we can remove the multiple retries and just allow the exception to bubble up and be caught, rather than attempt to work out which write operation is correct

src/Uno.Extensions.Configuration/Reloader.cs
```diff
@@ -32,10 +34,21 @@ public async Task ReloadAllFileConfigurationProviders(string? configFile = defau
...
 				{
-					var contents = File.ReadAllText(info.PhysicalPath);
-					if (Logger.IsEnabled(LogLevel.Debug)) Logger.LogDebugMessage($@"Contents '{contents}'");
-					if (Logger.IsEnabled(LogLevel.Debug)) Logger.LogDebugMessage($@"Loading from full path '{info.PhysicalPath}'");
-					fp.Load();
+					if (Logger.IsEnabled(LogLevel.Debug))
+					{
+						var contents = File.ReadAllText(info.PhysicalPath);
+						Logger.LogDebugMessage($@"Contents '{contents}'");
+						Logger.LogDebugMessage($@"Loading from full path '{info.PhysicalPath}'");
```
jeromelaban, line 41: This will become a performance issue, as the formatting is done regardless of the logging level

jeromelaban (approved):

src/Uno.Extensions.Configuration/WritableOptions.cs
carldebilly: I agree the lock is useless and this should never fail on mobile applications.  But on "real" platforms (Mac, Windows, Linux...) it could lead to pretty serious concurrency issues and losing information.

This is not ACID compliant and warning should be added in the documentation about this.

src/Uno.Extensions.Configuration/WritableOptions.cs
```diff
@@ -61,7 +61,15 @@ public async Task UpdateAsync(Func<T, T> applyChanges)
...
-		File.WriteAllText(physicalPath, json);
+		await Reloader.ReadWriteLock.WaitAsync();
+		try
+		{
+			File.WriteAllText(physicalPath, json);
+		}
+		finally
+		{
+			Reloader.ReadWriteLock.Release();
+		}
```
carldebilly, line 72: This is potentially dangerous code: if the reported problem what a sharing violation, that mean there was concurrent access to the file.

The new code will simply silently forget one of the concurrent change, which is a very dangerous pattern.

The fix for that would be to re-read the value and reapply the changes when a sharing violation is occurring.

carldebilly (changes requested):

src/Uno.Extensions.Configuration/WritableOptions.cs
dr1rrb: At least the lock should be around the whole update method.

src/Uno.Extensions.Configuration/WritableOptions.cs
nickrandolph: Updated core removes multiple retries but leaves the locks - if the read/write then fails, this bubbles up an exception as you would expect

src/Uno.Extensions.Configuration/WritableOptions.cs
nickrandolph: I suspect it depends on platform but my understanding is that sharing violation can happen for concurrent reads and writes. This matches where the exception was being raised which is coming out of the code that's attempting to reload the configuration information (ie reads).
Wrapping the rest of the logic of the update method is likely to cause deadlocks, given there's a call to the update callback. The lock should just cover the actual read/write of the file - if there's still an exception, this will bubble up to the calling method. This is no different from the FIFO behaviour you would expect from this code if there was no lock

nickrandolph: Initially I had left the lock around just the write as I was concerned about calling applyChanges (ie external code) within the lock. I've reorganised the method to invokve the applyChanges first and then wrap the whole read/write operation in the lock, as suggested.

francoistanguay (approved):

carldebilly (approved):

PR #788, opened 2022-10-03
Core code gen

dr1rrb:
## Refactoring
Move key equality and compat types generators to a dedicated Uno.Extensions.Core.Generator assembly

* We are using linked files for the compatibility types in extensions solution
* Key equality is generated by the Reactive package

## PR Checklist


src/Uno.Extensions.Core.Generators/Rules.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace Uno.Extensions.Reactive.Generator;
+
+[SuppressMessage("ReSharper", "InconsistentNaming")]
```
nickrandolph, line 9: Are ReSharper naming rules part of our coding standard? If not, we should remove this SuppressMessage; if they are, we should be enforcing this via the build

src/Uno.Extensions.Core.Generators/Uno.Extensions.Core.Generators.csproj
```diff
@@ -0,0 +1,50 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>netstandard2.0</TargetFramework>
+		<IsPackable>false</IsPackable>
+		<RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
```
nickrandolph, line 6: Are RuntimeIdentifiers required?

src/Uno.Extensions.Core.Generators/Uno.Extensions.Core.Generators.csproj
```diff
@@ -0,0 +1,50 @@
...
+		
+				<!--
+			Defines that this project output is a tool of the given package ID,
+			so nuget override will copy it in the 'tools' directory of that package instead of the 'lib' of the current package folder.
+		-->
+		<ToolOfPackage>Uno.Extensions.Reactive</ToolOfPackage>
```
nickrandolph, line 12: Is this still required? If so, it feels weird that this is Uno.Extensions.Core.Generators, instead of specific to Reactive

src/Uno.Extensions.Reactive.UI/AssemblyInfo.cs
```diff
@@ -1,3 +1,6 @@
 ﻿using System.Runtime.CompilerServices;
+using Uno.Extensions.Equality;
 
 [assembly: InternalsVisibleTo("Uno.Extensions.Reactive.Tests")]
```
nickrandolph, line 4: Can we move these properties into the csproj and then remove the AssemblyInfo.cs?

src/Uno.Extensions.Core.Generators/Uno.Extensions.Core.Generators.csproj
dr1rrb: Yes it's required, but not the right value ... changed it!

src/Uno.Extensions.Core.Generators/Rules.cs
dr1rrb: Nope, didn't saw it, removed!

src/Uno.Extensions.Reactive.UI/AssemblyInfo.cs
dr1rrb: I saw both pattern in the solution ... not sure which one I prefer. IMO the property in *.csproj is more for "generic" (like `$(AssemblyName).Tests`, but this does not apply here since we have only one tests project for all reactive packages for instance). And the advantage of the attribute / AssemblyInfo.cs file is to have a single place where we do have all those kind of configurations for the assembly (like the attribute I added).

BTW I also think that we should globally review usage of the `InternalsVisibleTo` in the solution and keep only those for tests projects.

src/Uno.Extensions.Core.Generators/Uno.Extensions.Core.Generators.csproj
dr1rrb: Well you add them in reactive package https://github.com/unoplatform/uno.extensions/pull/348 but I don't known why. I kept the same config to reduce risk of regression, but I agree that I don't see why there are needed. Why did you add them by that time @nickrandolph ?

nickrandolph: Unless I'm mistaken I think we've removed the RuntimeIdentifiers from both Core and Reactive projects - they were needed at some point for things to build but no longer an issue afaik for netstandard projects

dr1rrb: Ok then we should remove them from all projects

nickrandolph (approved):

PR #842, opened 2022-10-19
chore: SoftInput.AdjustNothing for droid template

kazo0:
With the Toolkit containing the [SafeArea](https://platform.uno/docs/articles/external/uno.toolkit.ui/doc/controls/SafeArea.html?tabs=none) control/behavior, we should change the `WindowSoftInputMode` on Android to `AdjustNothing` so that the system does not clash with the SafeArea's logic when the [SoftInput](https://platform.uno/docs/articles/external/uno.toolkit.ui/doc/controls/SafeArea.html?tabs=none#using-insetmasksoftinput-for-on-screen-keyboards) flag is set on SafeArea.


nickrandolph (approved):

nickrandolph:
@kazo0 I've merged this PR. Did you want to add another PR to include the SafeArea control across the two pages in the template?

kazo0:
@nickrandolph yeah that's probably a good idea, maybe just like a dummy login page to show how it might work? I'll make a PR

PR #917, opened 2022-11-14
feat(reactive): Allow generation of VM on non UI assembly

dr1rrb:

## Feature
Allow generation of VM on non UI assembly

Generated code depends on Uno / UWP / WinUI's dispatcher

## What is the new behavior?
The generated code non longer depends on dispatcher

## PR Checklist
- [x] Tested code with current [supported SDKs](../README.md#supported)


src/Directory.Build.props
```diff
@@ -171,10 +171,4 @@
 
 	<Import Project="Uno.CrossTargeting.props" />
 
-	<ItemGroup>
-		<InternalsVisibleTo Include="$(AssemblyName).UI" />
```
nickrandolph, line 175: Why have these internalsvisibleto items been removed?

src/Uno.Extensions.Core/IDispatcher.cs
```diff
@@ -1,18 +1,32 @@
-﻿namespace Uno.Extensions;
+﻿using System.ComponentModel;
```
nickrandolph, line 1: Push to GlobalUsings

src/Uno.Extensions.Core/IDispatcher.cs
```diff
@@ -1,18 +1,32 @@
...
+	bool HasThreadAccess { get; }
+
+	/// <summary>
+	/// Adds a task to the queue which will be executed on the thread associated with the dispatcher.
+	/// </summary>
+	/// <remarks>This is the raw version which allows to interact with the native dispatcher the fewest overhead possible.</remarks>
+	/// <param name="action">The task to execute.</param>
+	/// <returns>True indicates that the task was added to the queue; false, otherwise.</returns>
+	[EditorBrowsable(EditorBrowsableState.Advanced)] // Applications should prefer to use the ExecuteAsync which allow to track the execution
+	bool TryEnqueue(Action action);
```
nickrandolph, line 22: This makes IDispatcher super confusing for developers. Can we expose on different interface or just internally?

src/Uno.Extensions.Core/Threading/FastAsyncLock.cs
```diff
@@ -0,0 +1,240 @@
+﻿// ******************************************************************
+// Copyright � 2015-2018 nventive inc. All rights reserved.
```
nickrandolph, line 2: Copyright?

src/Uno.Extensions.Navigation.UI/Dispatcher.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Windows.UI.Xaml;
+﻿using Windows.UI.Core;
+using Windows.UI.Xaml;
```
nickrandolph, line 2: move to globalusing

src/Directory.Build.props
dr1rrb: Because they are causing lot of troubles with "compiler compatibility classes" like `ModuleInitializerAttribute`: When we reference 2 projects that are defining them, the compiler is not able to choose one.

IMO We should almost never use the "InternalsVisibleTo", even for tests, if we are testing specific internal implementation, I started to prefer reflection over it.

src/Uno.Extensions.Core/IDispatcher.cs
dr1rrb: Well I disagree : this is the contract that developers are expecting when they are used to deal with the native dispatcher. 

The issue with the default method is that it forces users to wrap every thing in a `Task` which can be costly for nothing. I would rather move the `ExecuteAsync` to be an extensions method.

BTW this method is flagged as `[EditorBrowsable(EditorBrowsableState.Advanced)]` so it won't be the first suggested.

src/Uno.Extensions.Core/Threading/FastAsyncLock.cs
dr1rrb: It has been copied from Uno.Core project https://github.com/unoplatform/Uno.Core/blob/master/src/Uno.Core.Extensions.Threading/FastAsyncLock.cs, as the file do have a copyright, I think that we should keep it.

carldebilly (approved):
TL;JA

PR #926, opened 2022-11-18
feat: adding IApplicationBuilder

dansiegel:

- #918

## PR Type

- Feature

All code revolves around the IHostBuilder

## What is the new behavior?

A new IApplicationBuilder has been introduced to help simplify the process of bootstrapping the application startup. Currently the ShowAsync has been moved to the template until we determine the best location for this as Toolkit Navigation doesn't actually have a reference to Hosting so it cannot access IApplicationBuilder. There is also no current way to pass the boolean for Uno logging when creating the Host.

## PR Checklist

## Other information


src/Uno.Extensions.Hosting.UI/ApplicationBuilder.cs
```diff
@@ -0,0 +1,41 @@
...
+	public IHost Build()
+	{
+		var builder = UnoHost.CreateDefaultBuilder();
+		foreach (var @delegate in _delegates)
+		{
+			@delegate(builder);
+		}
+
+		// TODO: Need to expose the "enableUnoLogging" as part of UseLogging
+		return builder.Build();
```
nickrandolph, line 33: We now have https://github.com/unoplatform/uno.extensions/blob/a076d451484ce7e46a0577f730f7bcfa0ff2a683/src/Uno.Extensions.Hosting/IServiceInitialize.cs#L6 which didn't exist when we first wrote the startup logic.
We should extend UseLogging to specify the "enableUnoLogging" option and then use that to register an ISerivceInitialize instance which will do the same logic as calling Build(enabledUnoLogging: true)

src/Uno.Extensions.Hosting.UI/IApplicationBuilder.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Microsoft.UI.Xaml;
```
nickrandolph, line 1: globalusing

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.UI/HostConfiguration.cs
```diff
@@ -0,0 +1,139 @@
+namespace MyExtensionsApp;
+
+public static class HostConfiguration
+{
+	// Register services for the application
+	private static void ConfigureServices(IServiceCollection services)
+	{
```
nickrandolph, line 7: Change this to be the same as the other extension methods on IApplicationBuilder

nickrandolph:
Looking good but a couple of outstanding things:
- Need to update Logging to provide implementation of IServiceInitialize for when unologging is enabled
- Need to alter InitializeNavigationAsync so that there's only one method and that if user calls UseToolkitNavigation, Navigation.Toolkit gets an opportunity to initialize either LoadingView or SplashScreen accordingly

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.UI/HostConfiguration.cs
dansiegel: actually the pattern I was following was to use the same name as the underlying method that was being called which in this case was Configure instead of Use... if you want to drop the Use then it would make sense to drop the configure

src/Uno.Extensions.Hosting.UI/ApplicationBuilder.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace Uno.Extensions.Hosting;
+
+internal class ApplicationBuilder : IApplicationBuilder
```
nickrandolph, line 3: Could this be a record?

src/Uno.Extensions.Hosting.UI/ApplicationBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+	public ApplicationBuilder(Application app, LaunchActivatedEventArgs arguments)
+	{
+		App = app;
+		Arguments = arguments;
+
+#if NET6_0_OR_GREATER && WINDOWS && !HAS_UNO
+		Window = new Window();
+#else
+        Window = Microsoft.UI.Xaml.Window.Current;
+#endif
```
nickrandolph, line 15: This code could be directly on instance property

src/Uno.Extensions.Hosting.UI/ApplicationBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	public Application App { get; }
+	public LaunchActivatedEventArgs Arguments { get; }
+	public Window Window { get; }
+
+	public IHost Build()
+	{
+		var builder = UnoHost.CreateDefaultBuilder();
+		foreach (var @delegate in _delegates)
```
nickrandolph, line 25: Avoid @ to force variable name - just give it a different name eg "del"

src/Uno.Extensions.Hosting.UI/ApplicationBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+		{
+			@delegate(builder);
+		}
+
+		// TODO: Need to expose the "enableUnoLogging" as part of UseLogging
```
nickrandolph, line 30: Calling out the TODO here to make sure it gets addressed: // TODO: Need to expose the "enableUnoLogging" as part of UseLogging

src/Uno.Extensions.Logging/HostBuilderExtensions.cs
```diff
@@ -6,13 +6,14 @@ public static class HostBuilderExtensions
 {
 	public static IHostBuilder UseLogging(
 		this IHostBuilder hostBuilder,
-		Action<ILoggingBuilder> configure)
+		Action<ILoggingBuilder> configure, bool enableUnoLogging = false)
```
nickrandolph, line 9: This is what we do for navigation:
https://github.com/unoplatform/uno.extensions/blob/081dad60b5d5f98906a6782f3ea687aa75100b5d/src/Uno.Extensions.Navigation.UI/HostBuilderExtensions.cs#L10
We have a NavigationConfig function callback - I would suggest we try to maintain consistency with this.

src/Uno.Extensions.Logging/Uno.Extensions.Logging.UWP.Skia.csproj
```diff
@@ -13,5 +13,9 @@
 		<PackageReference Include="Uno.UI" />
 	</ItemGroup>
 
+	<ItemGroup>
```
nickrandolph, line 16: Move to common.props

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.UI/HostConfiguration.cs
nickrandolph: Yes, agreed, drop Configure here too

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.UI/HostConfiguration.cs
```diff
@@ -0,0 +1,138 @@
...
+			splashScreen.Initialize(appBuilder.Window, appBuilder.Arguments);
+		}
+
+		appBuilder.Window.Content = appRoot;
+		appBuilder.Window.Activate();
+
+		// TODO: This needs tidying up inside Navigation library so that there's a single
+		// InitializeNavigationAsync method that can make the determination on whether
+		// to use the logic from Navigation.Toolkit or not
+		if (navRoot is LoadingView loading)
```
nickrandolph, line 115: Two things that need resolving so that we can move ShowAsync into extensions library:
- we need a single InitializeNavigationAsync method that can make the determination on whether to use the logic from Navigation.Toolkit or not
- we need a way that the InitializeNavigationAsync parameter can be specified - allows for dynamic determination of first navigation request (for example if user is logged in go to home page, otherwise login page)

src/Uno.Extensions.Hosting.UI/ApplicationBuilder.cs
nickrandolph: Let's remove this TODO

nickrandolph (approved):

Issue #927, opened 2022-11-18
[Storage] Content files that are in a sub-folder aren't accessible on Android


nickrandolph:
## Current behavior

If you have a Content file in a sub-folder (eg /Assets/MyPhoto.png ) you can't access it via IStorage interface on Android. This is because of the logic added to cache the index of files which only retrieves the list at the root folder level.

## Expected behavior

Content in sub-folders should be accessible

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renders for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

jeromelaban:
Asset files are only accessible through the https://platform.uno/docs/articles/features/file-management.html?q=storage#support-for-storagefilegetfilefromapplicationuriasync. Is this how you're getting the files?

PR #956, opened 2022-11-28
fix: Sub-folder content files aren't accessible on Android

arqueror:

https://github.com/unoplatform/uno.extensions/issues/927

## PR Type

- Fix

Content files that are in a sub-folder aren't accessible on on FileExistsInPackage() logic for Android

## What is the new behavior?
Added Android specific code to FileStorage.cs so it is able to scan package assets/resources correctly

## Environment
Affected platform(s):

- [ ] iOS
- [x] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

## PR Checklist


src/Uno.Extensions.Storage.UI/FileStorage.cs
```diff
@@ -53,36 +47,60 @@ public async Task<string> CreateFolderAsync(string foldername)
 	{
 		try
 		{
+#if __ANDROID__
+				var assets = global::Android.App.Application.Context.Assets;
+				var inputStream = assets?.Open(filename);
+				var content = inputStream?.ReadToEnd();
+				inputStream?.Close();
```
jeromelaban, line 54:
```suggestion
				using var inputStream = assets?.Open(filename);
				var content = inputStream?.ReadToEnd();
				inputStream?.Close();
```

jeromelaban, line 54: It's likely not needed to close explicitly since disposing it is good enough.

src/Uno.Extensions.Storage.UI/FileStorage.cs
```diff
@@ -53,36 +47,60 @@ public async Task<string> CreateFolderAsync(string foldername)
...
 	}
 
 	public async Task<Stream?> OpenPackageFileAsync(string filename)
 	{
-		if (!await FileExistsInPackage(filename))
+
+		try
+		{
+#if __ANDROID__
+			var assets = global::Android.App.Application.Context.Assets;
```
jeromelaban, line 85: You should not need to handle this this way, as `GetFileFromApplicationUriAsync` has support for it. If it's not working, then we need to fix uno (as it uses the same APIs underneath).

jeromelaban: Try creating a repro case using the vanilla uno template (a simple file/new uno project) and see if you can repro the same issue.

jeromelaban: See https://github.com/unoplatform/uno/blob/0f612951b0c97fd00653ff9cd68ccddd420eefed/src/Uno.UI.RuntimeTests/Tests/Windows_Storage/Given_ApplicationStorage.cs for runtime tests about this feature, and https://github.com/unoplatform/uno/blob/b6c132e7520307c2f7560b373ce858e5d1a16f12/src/Uno.UWP/Storage/StorageFile.Android.cs#L25 for the code for this feature.

It's best that you close this PR and open an issue in unoplatform/uno, then try to fix the issue there.

nickrandolph: Afaik there's no issue with behaviour of Uno storage APIs - it should correctly throw exception if attempting to open a file that doesn't exist. This is uwp/winui behaviour (even if I disagree with it). 
The point of the Android specific code here is to do File.exists check for application packaged files. Rather than removing the existing block of code, it needs to be extended to handle sub-folders.

arqueror: @nickrandolph i discovered there is an issue with the UNO Storage Api on Android by creating a new project. If you want to load let's say "Fonts/uno-fluentui-assets.ttf" or "AboutAssets.txt"which both are in there by default, logic is failing to retrieve it because of line 39 which is modifying some chars in the file name:

`var path = AndroidResourceNameEncoder.EncodeResourcePath(originalPath);`

For the File.Exists i agree, it should be the only change added in this PR. Extend the FIle.Exist native check part and then try to load it with `StorageFile.GetFileFromApplicationUriAsync ` just as it is right now.

jeromelaban: Indeed, this particular line should only be used to resolve resources, not assets. Can you try to make a PR that adjusts this behavior, with additional tests?

arqueror: In the end i just realized that i don't think we need this check for android since the storage Api base class is already making this validation and even further than that, is able to make a distinction of if what you're looking for is a resource/asset. The only difference in reality is that extensions wrapper is just wrapping the call with a try/catch for handling errors and then returning a null but nothing more than that.

I found out as well that the process for looking for an asset vs resource is completelly different in ANdroid and not as straight-forward as in Win and MacOS|iOS, and in fact the logic we have in place now in the native check will only return non drawable files(assets) and base class is already handling this(asets and resources). I Think Jerome added support for resources recently.Already tested without the native check (just returning true) and its working as expected when reading from root/nested paths. Tested with runtime tests as well and it's working as expected.

src/Uno.Extensions.Storage.UI/FileStorage.cs
```diff
@@ -11,10 +13,45 @@ public class FileStorage : IStorage
...
 	{
 #if __ANDROID__
-		var assets = global::Android.App.Application.Context.Assets;
-		var files = assets?.List("");
-		filename = Path.GetFileNameWithoutExtension(filename).Replace('.', '_') + Path.GetExtension(filename);
-		return files?.Contains(filename)??false;
+		//Look in assets first***
+		var context = global::Android.App.Application.Context;
+		var assets = context.Assets;
+		var normalizedFileName = Path.GetFileNameWithoutExtension(filename).Replace('.', '_').Replace('-','_') + Path.GetExtension(filename);
```
nickrandolph, line 19: Can we use the EncodeResourcePath from https://github.com/unoplatform/uno/blob/43ca40a9f9c86160d85fe346fe74682c2ae5da4a/src/Uno.UWP/Helpers/AndroidResourceNameEncoder.cs#L53
cc @jeromelaban would it be worth exposing this method so it can be used by extensions etc?

jeromelaban: That's a good point. I think we should be moving this code to uno, as an extension, as it relies on internal implementation details about the way resources are encoded.

Having something like `Uno.Storage.StorageFileHelper.Exists()` would be interesting instead.

PR #957, opened 2022-11-29
fix: ShowAsync should be part of Hosting

dansiegel:

n/a

## PR Type

- Bugfix

ShowAsync is a parameterless option for the NavigateAsync extension in the Navigation library

## What is the new behavior?

NavigateAsync is available with/without specifying an initial navigation callback.
ShowAsync is an extension in the Hosting Library which simply sets the Shell and initializes the host.

### DIFF
--- a/samples/Playground/Playground.Shared/App.xaml.cs
+++ b/samples/Playground/Playground.Shared/App.xaml.cs
@@ -148,7 +148,7 @@ public sealed partial class App : Application
 
 			case InitOption.AppBuilderShell:
 
-				_host = await appBuilder.ShowAsync<AppRoot>();
+				_host = await appBuilder.NavigateAsync<AppRoot>();
 				break;
 
 			case InitOption.NoShellViewModel:
--- a/src/Uno.Extensions.Hosting.UI/ApplicationExtensions.cs
+++ b/src/Uno.Extensions.Hosting.UI/ApplicationExtensions.cs
@@ -1,6 +1,4 @@
-﻿using Microsoft.UI.Xaml;
-
-namespace Uno.Extensions.Hosting;
+﻿namespace Uno.Extensions.Hosting;
 
 /// <summary>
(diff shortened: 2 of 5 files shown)


src/Uno.Extensions.Hosting.UI/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+public static class ApplicationBuilderExtensions
+{
+	/// <summary>
+	/// Creates a new Shell for the main content of the Window, builds the <see cref="IHost" />
+	/// and starts the Host before finally returning it.
+	/// </summary>
+	/// <typeparam name="TShell">The <see cref="UIElement" /> to set as the <see cref="Window" /> content.</typeparam>
+	/// <param name="builder">The <see cref="IApplicationBuilder" />.</param>
+	/// <returns>The started <see cref="IHost" />.</returns>
+	public static async Task<IHost> ShowAsync<TShell>(this IApplicationBuilder builder)
```
nickrandolph, line 15: This shouldn't exist in Hosting, well not in its current form. It's unlikely that this would ever be used as is. Most common (other than using navigation) would be to create frame and nav to first page, before activating window.

On this topic, we need to review where Activate is called in the showasync and navigateasync methods to ensure nav is complete before Activate is called.

nickrandolph (changes requested):
Need to revisit this. Nothing about ShowAsync should exist in Hosting, which is fundamentally about creating the host of the app. If Devs aren't using navigation, we should be building IHost and leaving it up to the developer to set root visual and call activate

src/Uno.Extensions.Hosting.UI/ApplicationBuilderExtensions.cs
dansiegel: removed...

nickrandolph (approved):

PR #975, opened 2022-12-08
Single Location Project & Server

dansiegel:
## PR Type

- Project Template Refactoring

Template uses a Shared Project

## What is the new behavior?

Template updated to newly supported Single Location Project format. 
Adds new Server project as an option for both an API and WASM host if the WASM head is added to the solution.
Refactors some property names and fixes ide.host.json so that certain parameters can be hidden from the UI. Adds new option for using the Http Extensions.

**NOTE**
- New DataContracts project will be created if either the Server or Http flags are true


src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Build.nocpm.props
```diff
@@ -9,9 +9,18 @@
 		<SynthesizeLinkMetadata>true</SynthesizeLinkMetadata>
 
 		<TreatWarningsAsErrors>True</TreatWarningsAsErrors>
-	  	<NoWarn>$(NoWarn);Uno0001;CS1998;CA1416;NU1507</NoWarn>
+		<NoWarn>$(NoWarn);Uno0001;CS1998;CA1416;NU1507</NoWarn>
 	</PropertyGroup>
 
+	<ItemGroup Condition=" $(UnoSingleLocationProject) == 'true' ">
```
nickrandolph, line 15: This feels sloppy - is this consistent with the Uno vanilla template, or is this specific to the extensions template

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Packages.props
```diff
@@ -22,26 +22,32 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageVersion Include="NUnit" Version="3.13.2" />
     <PackageVersion Include="NUnit3TestAdapter" Version="4.0.0" />
+    <!--#if (server && use-serilog)-->
```
nickrandolph, line 25: Do we really need all these conditional flags in the Packages.props - all this is doing is setting version number for when packages are used in cpm? I get that this keeps these files clean but I'm just thinking of the overhead of maintaining more conditional flags

src/Uno.Extensions.Templates/Uno.Extensions.Templates.csproj
```diff
@@ -19,55 +19,75 @@
...
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//PackageVersion[@Version='255.255.255.255']/@Version" Value="$(PackageVersion)" Namespaces="" />
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//PackageReference[@Version='255.255.255.255']/@VersionOverride" Value="$(PackageVersion)" Namespaces="" />
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//PackageReference[@Version='255.255.255.255']/@Version" Value="$(PackageVersion)" Namespaces="" />
+	</Target>
 
+	<Target Name="ResetVersions" AfterTargets="Pack">
 
+		<ReplaceFileText Filename="%(_AllCSProjFiles.Identity)"
+					MatchExpression="Version=&quot;$(PackageVersion)&quot;"
+					ReplacementText="Version=&quot;255.255.255.255&quot;" />
```
nickrandolph, line 66: Aren't these values the wrong way around - aren't we matching on 255.255.255.255?

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Packages.props
```diff
@@ -19,29 +19,35 @@
     <!--#endif-->
     <PackageVersion Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.755" />
     <PackageVersion Include="Microsoft.WindowsAppSDK" Version="1.2.221116.1" />
```
nickrandolph, line 21: Can we update this to [1.2.221209.1](https://www.nuget.org/packages/Microsoft.WindowsAppSDK/1.2.221209.1)

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Packages.props
```diff
@@ -55,25 +61,36 @@
...
+    <PackageVersion Include="Uno.Themes.WinUI.Markup" Version="2.4.0-dev.68" />
     <!--#endif-->
-    <PackageVersion Include="Uno.WinUI.RemoteControl" Version="4.6.19" />
-    <PackageVersion Include="Uno.WinUI.Skia.Gtk" Version="4.6.19" />
-    <PackageVersion Include="Uno.WinUI.Skia.Linux.FrameBuffer" Version="4.6.19" />
-    <PackageVersion Include="Uno.WinUI.Skia.Wpf" Version="4.6.19" />
-    <PackageVersion Include="Uno.WinUI.WebAssembly" Version="4.6.19" />
-    <PackageVersion Include="Xamarin.Google.Android.Material" Version="1.4.0.4" />
-    <PackageVersion Include="Xamarin.UITest" Version="3.2.6" />
+    <PackageVersion Include="Uno.WinUI.RemoteControl" Version="4.7.0-dev.564" />
```
nickrandolph, line 88: Are we still pushing to single location? If not, can we revert Uno refs to latest stable

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp-vsmac.slnf
```diff
@@ -9,10 +9,16 @@
       "MyExtensionsApp.Skia.Gtk\\MyExtensionsApp.Skia.Gtk.csproj",
 #//#endif
 #//#if (wasm)
-      "MyExtensionsApp.Wasm\\MyExtensionsApp.Wasm.csproj",
+      "MyExtensionsApp.Skia.Wasm\\MyExtensionsApp.Wasm.csproj",
```
nickrandolph, line 12: Skia.Wasm?

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.DataContracts/MyExtensionsApp.DataContracts.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<TargetFramework>net6.0</TargetFramework>
+		<Nullable>enable</Nullable>
```
nickrandolph, line 4: this is in the root Directory.Build.props

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.DataContracts/MyExtensionsApp.DataContracts.csproj
```diff
@@ -0,0 +1,8 @@
...
+		<GenerateDocumentationFile>true</GenerateDocumentationFile>
+		<NoWarn>$(NoWarn);CS1591</NoWarn>
```
nickrandolph, line 6: We should include comment as to why this issue is ignored

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.DataContracts/WeatherForecast.cs
```diff
@@ -0,0 +1,15 @@
...
+/// </summary>
+/// <param name="Date">Gets the Date of the Forecast.</param>
+/// <param name="TemperatureC">Gets the Forecast Temperature in Celsius.</param>
+/// <param name="Summary">Get a description of how the weather will feel.</param>
+public record WeatherForecast(DateTime Date, int TemperatureC, string? Summary)
+{
+	/// <summary>
+	/// Gets the Forecast Temperature in Fahrenheit
+	/// </summary>
+	public int TemperatureF => 32 + (int)(TemperatureC / 0.5556);
```
nickrandolph, line 14: I'm not sure why this uses a static constant when this should be 5/9

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Mobile/Android/AndroidManifest.xml
```diff
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.companyname.myextensionsapp" android:versionCode="1" android:versionName="1.0">
-  <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="31" />
-  <application android:label="MyExtensionsApp"></application>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <application android:allowBackup="true" android:supportsRtl="true"></application>
```
nickrandolph, line 3: I think it might be useful to have a comment here indicating where min sdk and target sdk version come from now

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Mobile/MacCatalyst/Info.plist
```diff
@@ -1,46 +1,29 @@
...
-	<key>UISupportedInterfaceOrientations</key>
-	<array>
-	  <string>UIInterfaceOrientationPortrait</string>
-	  <string>UIInterfaceOrientationLandscapeLeft</string>
-	  <string>UIInterfaceOrientationLandscapeRight</string>
-	</array>
-	<key>XSAppIconAssets</key>
-	<string>Media.xcassets/AppIcon.appiconset</string>
-	<key>UIAppFonts</key>
-	<array>
```
nickrandolph, line 32: Where are fonts added now?

src/Uno.Extensions.Templates/Uno.Extensions.Templates.csproj
dansiegel: it replaces the 255.255.255.255 post build

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Packages.props
dansiegel: no we will not be targeting latest stable as there should still be some updated targets in the 4.7 dev build... tbd once the updated template format is finalized

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.DataContracts/WeatherForecast.cs
dansiegel: The implementation was pulled from the Uno template and the .NET Team's API template and just updated for your love of records

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Mobile/MacCatalyst/Info.plist
dansiegel: These should be added by Uno now

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Mobile/Android/AndroidManifest.xml
dansiegel: This comes from the Android SDK and the version you specify in your csproj

dansiegel:
Outstanding issues:

1. For WinUI head, when including `MaterialToolkitResources` in the merged dictionaries a Microsoft.UI.Xaml.UnhandledExceptionEventArgs is thrown with the message `Element is already the child of another element.`
2. For WinUI head an exception is thrown when calling `builder.Build` or invoking `builder.NavigateAsync` which internally calls the Build. 
    ```
    Exception of type 'System.InvalidOperationException' was thrown.
       at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|20_0(Int32 hr)
       at ABI.Windows.Storage.IApplicationDataStaticsMethods.get_Current(IObjectReference _obj)
       at Uno.Extensions.Hosting.UnoHost.<>c.<CreateDefaultBuilder>b__0_0(HostBuilderContext ctx, IConfigurationBuilder appConfig)
       at Uno.Extensions.Hosting.HostBuilder.BuildAppConfiguration()
       at Uno.Extensions.Hosting.HostBuilder.Build()
       at Uno.Extensions.Hosting.ApplicationBuilder.Build()
       at Uno.Extensions.ApplicationBuilderExtensions.<>c__DisplayClass0_0`1.<NavigateAsync>b__1()
       at Uno.Extensions.ServiceProviderExtensions.<BuildAndInitializeHostAsync>d__17.MoveNext()
       at Uno.Extensions.ApplicationBuilderExtensions.<NavigateAsync>d__0`1.MoveNext()
       at UnoExtApp88.App.<OnLaunched>d__3.MoveNext() 
    ```
3. For the WinUI head when creating a blank app (and this likely is present but hidden on the default template due to the other bugs) the following exception is thrown when Navigating to the MainPage
    ```
    Microsoft.UI.Xaml.Markup.XamlParseException: 'XAML parsing failed.'
       at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|20_0(Int32 hr)
       at ABI.Microsoft.UI.Xaml.IApplicationStaticsMethods.LoadComponent(IObjectReference _obj, Object component, Uri resourceLocator, ComponentResourceLocation componentResourceLocation)
       at UnoExtApp89.MainPage.InitializeComponent()
    ```

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -1,22 +1,22 @@
...
+    "Mac Catalyst",
+    "Windows",
     "Linux",
-    "WebAssembly",
-    "WinUI",
-    "WindowsAppSdk",
-    "Extensions",
-    "dotnet-new"
+    "WebAssembly"
+    // "WinUI",
```
nickrandolph, line 14: We need to keep WinUI, since this is used by both WinAppSdk and TemplateStudio - we want uno templates to show up in this category

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -1,22 +1,22 @@
...
+    // "WindowsAppSdk",
+    // "Extensions"
   ],
   "name": "Multi-Platform App (Uno.Extensions)",
-  "identity": "Uno.Extensions",
+  "identity": "Uno.Extensions.TemplateStudio.CSharp",
```
nickrandolph, line 19: Remove TemplateStudio, so just Uno.Extensions.CSharp

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -55,47 +78,90 @@
...
           "choice": "mvvm",
           "description": "Use the MVVM design pattern",
-          "displayName": "MVVM"
+          "displayName": "MVVM (Model-View-ViewModel)"
         },
         {
           "choice": "mvux",
-          "description": "Use the MVUx design pattern with Uno.Extensions.Reactive",
-          "displayName": "MVUx"
+          "description": "Use the MVU-X design pattern with Uno.Extensions.Reactive",
```
nickrandolph, line 113: Remove "with Uno.Extensions.Reactive"

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -266,22 +399,89 @@
...
+      "type": "generated",
+      "generator": "switch",
+      "replaces": "$libarybasetargetframework$",
+      "parameters": {
+        "evaluator": "C++",
+        "datatype": "string",
+        "cases": [
+          {
+            "condition": "(tfm == 'net6.0')",
+            "value": "netstandard2.1"
```
nickrandolph, line 412: Why wouldn't we default back to net6.0?

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -266,22 +399,89 @@
...
+    "HostIdentifier": {
+      "type": "bind",
+      "binding": "HostIdentifier"
     }
   },
   "primaryOutputs": [
     {
-      "path": "MyExtensionsApp.sln",
-      "condition": "!is-vsix"
+      "condition": "(HostIdentifier == \"dotnetcli\" || HostIdentifier == \"dotnetcli-preview\" || HostIdentifier == \"vs\")",
```
nickrandolph, line 472: When is this not true?

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -397,12 +589,60 @@
             "MyExtensionsApp.Skia.Linux.FrameBuffer/**/*"
           ]
         },
+        {
+          "condition": "(winAppSdk && skia-gtk)",
```
nickrandolph, line 593: why "winappsdk"?

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -397,12 +589,60 @@
...
+          "exclude": [
+            "MyExtensionsApp.Skia.Gtk/Package.appxmanifest",
```
nickrandolph, line 595: Shouldn't this just exclude the whole skia.gtk folder?

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -231,30 +327,67 @@
...
+      "type": "computed",
+      "datatype": "bool",
+      "value": "logging"
+    },
+    "use-serilog": {
+      "type": "computed",
+      "datatype": "bool",
+      "value": "(logging && serilog)"
+    },
+    "reactive": {
```
nickrandolph, line 382: For consistency, should we change this to  "use-reactive" or "use-mvux"?

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
nickrandolph: If we use net6.0 we can remove the compat files too

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -397,12 +589,60 @@
...
+          ]
+        },
+        {
+          "condition": "(!use-http && !server)",
+          "exclude": [
+            "MyExtensionsApp.DataContracts/**/*"
+          ]
+        },
+        {
+          "condition": "(architecture == 'mvvm')",
```
nickrandolph, line 625: change to use calculated symbol "use-mvvm"

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -459,29 +700,76 @@
...
+            "MyExtensionsApp/Styles/*.cs"
           ]
         },
         {
-          "condition": "(markup)",
+          "condition": "(use-csharp-markup)",
           "exclude": [
-            "MyExtensionsApp.UI/**/Styles/*.xaml",
-            "MyExtensionsApp.UI/**/Views/*.xaml"
+            "MyExtensionsApp/Styles/*.xaml",
```
nickrandolph, line 712: This line isn't required

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -459,29 +700,76 @@
...
+          "condition": "(default-app-template && navigation != 'frame')",
+          "exclude": [
+            "MyExtensionsApp/MainPage.xaml",
+            "MyExtensionsApp/MainPage.xaml.cs"
+          ]
+        },
+        {
+          "condition": "(tfm == 'net7.0')",
+          "exclude": [
+            "MyExtensionsApp.DataContracts/_Compat/**"
```
nickrandolph, line 764: See earlier comment about using net6.0 - we should be able to remove the _Compact folder completely

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Build.nocpm.props
```diff
@@ -1,6 +1,6 @@
 <Project ToolsVersion="15.0">
 	<PropertyGroup>
-		<LangVersion>10.0</LangVersion>
+		<LangVersion>latest</LangVersion>
```
nickrandolph, line 3: Change to latest known good version, which is 11. Concern with using latest is that it may cause issues if there are unsupported features introduced

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Build.nocpm.props
```diff
@@ -9,7 +9,56 @@
...
 
 		<TreatWarningsAsErrors>True</TreatWarningsAsErrors>
-	  	<NoWarn>$(NoWarn);Uno0001;CS1998;CA1416;NU1507</NoWarn>
+		<NoWarn>$(NoWarn);Uno0001;CS1998;CA1416;NU1507</NoWarn>
+
+		<DefaultLanguage>en</DefaultLanguage>
+
+		<!--#if (android)-->
+		<IsAndroid>false</IsAndroid>
+		<IsAndroid Condition="$(TargetFramework) != 'net6.0' AND $(TargetFramework) != 'netstandard2.1' AND $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">true</IsAndroid>
```
nickrandolph, line 18: Why do we need the tests against net6.0 and netstandard2.1? Isn't targetframework test sufficient?

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Build.nocpm.props
```diff
@@ -9,7 +9,56 @@
 		<SynthesizeLinkMetadata>true</SynthesizeLinkMetadata>
 
 		<TreatWarningsAsErrors>True</TreatWarningsAsErrors>
-	  	<NoWarn>$(NoWarn);Uno0001;CS1998;CA1416;NU1507</NoWarn>
+		<NoWarn>$(NoWarn);Uno0001;CS1998;CA1416;NU1507</NoWarn>
+
+		<DefaultLanguage>en</DefaultLanguage>
+
+		<!--#if (android)-->
```
nickrandolph, line 16: I think we can remove the conditions around these properties - if targets are added later, things will just work

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Packages.props
```diff
@@ -1,47 +1,43 @@
...
-    <PackageVersion Include="FluentAssertions" Version="6.7.0" />
-    <!--#if (msftextv6) -->
-    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="6.0.0" />
-    <!--#endif-->
-    <!--#if (msftextv7) -->
-    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
-    <!--#endif-->
+    <PackageVersion Include="coverlet.collector" Version="3.2.0" />
+    <PackageVersion Include="FluentAssertions" Version="6.8.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="$msftextlibversion$" />
```
nickrandolph, line 6: This doesn't need to be linked to tfm, can just use 7.0.0

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Packages.props
```diff
@@ -1,47 +1,43 @@
...
     <PackageVersion Include="Uno.Extensions.Hosting.WinUI" Version="255.255.255.255" />
     <PackageVersion Include="Uno.Extensions.Http" Version="255.255.255.255" />
     <PackageVersion Include="Uno.Extensions.Http.Refit" Version="255.255.255.255" />
-    <!--#if (localization)-->
+    <!--#if (use-localization)-->
     <PackageVersion Include="Uno.Extensions.Localization" Version="255.255.255.255" />
     <PackageVersion Include="Uno.Extensions.Localization.WinUI" Version="255.255.255.255" />
     <!--#endif-->
-    <!--#if (uselogging)-->
+    <!--#if (use-logging)-->
```
nickrandolph, line 38: Missing "endif"

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.blank.cs
```diff
@@ -0,0 +1,172 @@
...
+public sealed partial class App : global::Microsoft.UI.Xaml.Application
+{
+	private Window? _window;
+
+	public App()
+	{
+		InitializeLogging();
+		this.InitializeComponent();
+
+#if HAS_UNO || NETFX_CORE
```
nickrandolph, line 13: Remove the wiring up of Suspending to clean up source code

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.default.cs
```diff
@@ -15,9 +13,11 @@ public App()
 	{
 		this.InitializeComponent();
 
+//-:cnd:noEmit
 #if HAS_UNO || NETFX_CORE
```
nickrandolph, line 17: Remove OnSuspending

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.default.cs
```diff
@@ -48,39 +52,89 @@ protected async override void OnLaunched(LaunchActivatedEventArgs args)
...
+#if use-serilog
 				.UseSerilog(consoleLoggingEnabled: true, fileLoggingEnabled: true)
 #endif
 #endif
-#if configuration
+#if use-configuration
 				.UseConfiguration(configure: configBuilder =>
 					configBuilder
-						.EmbeddedSource<App>()
+						.EmbeddedSource<MainPage>()
```
nickrandolph, line 62: Change MainPage back to App

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.default.cs
```diff
@@ -111,7 +178,27 @@ private static void RegisterRoutes(IViewRegistry views, IRouteRegistry routes)
...
+			new ViewMap(ViewModel: typeof(ShellViewModel)),
+			new ViewMap<MainPage, MainViewModel>(),
+			new DataViewMap<SecondPage, SecondViewModel, Entity>()
+		);
+
+		routes.Register(
+			new RouteMap("", View: views.FindByViewModel<ShellViewModel>(),
+				Nested: new RouteMap[]
+				{
+					new RouteMap("Main", View: views.FindByViewModel<MainViewModel>()),
```
nickrandolph, line 195: If we change this to use FindByView instead of FindByViewModel then we don't need different code for reactive v mvvm

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.DataContracts/WeatherForecast.cs
nickrandolph: Just because MS shipped something that's not correct...... let's fix this please

src/Uno.Extensions.Templates/Uno.Extensions.Templates.csproj
```diff
@@ -19,55 +19,75 @@
...
 		<UpToDateCheckInput Include="content\**" />
+		<Compile Remove="content/**/*" />
+		<Page Remove="content/**/*" />
+		<Content Remove="content/**/*" />
+		<Content Include="content/**/*"
+				Exclude="**/obj/**;**/bin/**;**/*.Debugging.sln"
+				KeepDuplicates="False"
+				PackagePath="content"
+				Pack="true" />
+		<Content Include="content/unoapp-extensions/MyExtensionsApp.Windows/app.manifest"
```
nickrandolph, line 38: Are these individual Content includes required?

src/Uno.Extensions.Templates/Uno.Extensions.Templates.csproj
```diff
@@ -19,55 +19,75 @@
...
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//x:PackageReference[@Version='255.255.255.255']/@VersionOverride" Value="$(PackageVersion)" Namespaces="$(MSBuildDeveloperNamespace)" />
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//x:PackageReference[@Version='255.255.255.255']/@Version" Value="$(PackageVersion)" Namespaces="$(MSBuildDeveloperNamespace)" />
-
-		<!-- SDK Style projects -->
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//PackageVersion[@Version='255.255.255.255']/@Version" Value="$(PackageVersion)" Namespaces="" />
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//PackageReference[@Version='255.255.255.255']/@VersionOverride" Value="$(PackageVersion)" Namespaces="" />
-		<XmlPoke XmlInputPath="%(_AllCSProjFiles.Identity)" Query="//PackageReference[@Version='255.255.255.255']/@Version" Value="$(PackageVersion)" Namespaces="" />
+	</Target>
 
+	<Target Name="ResetVersions" AfterTargets="Pack">
```
nickrandolph, line 62: If someone does a build but not a pack, will this cause issues with the version being incorrectly updated?

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -174,55 +240,85 @@
...
     },
-    "configuration": {
+    "server": {
+      "displayName": "Server - API",
+      "type": "parameter",
+      "datatype": "bool",
+      "defaultValue": "true",
+      "description": "Includes a Server project for an API & host for the WebAssembly project"
+    },
+    "use-configuration": {
```
nickrandolph, line 250: We need to have consistency with the naming of these symbols. My preference is to drop the "use-", or perhaps only have the "use-" where the symbol is computed.

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/dotnetcli.host.json
```diff
@@ -1,6 +1,10 @@
 ﻿{
   "$schema": "http://json.schemastore.org/dotnetcli.host",
   "symbolInfo": {
+    "template-type": {
+      "longName": "template-type",
+      "shortName": "type"
+    },
     "appid": {
       "longName": "app-id",
```
nickrandolph, line 9: Can we default to having the longName be the same as the symbol name?

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
dansiegel: good catch that wasn't actually supposed to be one that I meant to remove

src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Packages.props
dansiegel: no the endif for this is on line 45. It wraps the `if serilog`

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.default.cs
dansiegel: It was changed to MainPage on purpose because it needs to actually work

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.default.cs
dansiegel: these are "duplicated" because of the difference in name between Reactive Models and MVVM ViewModels

src/Uno.Extensions.Templates/Uno.Extensions.Templates.csproj
dansiegel: Yes they are required. If you look at the package path it duplicates them from the Windows to the Linux / GTK heads

src/Uno.Extensions.Templates/Uno.Extensions.Templates.csproj
dansiegel: I mean to be fair the weren't being reset before... but I can add AfterBuild to the AfterTargets so that it will reset even if Pack isn't invoked but will continue to execute at the right point when we do pack

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
dansiegel: no the condition excludes duplicated files that are referenced from the WinAppSdk project but only when you include WinAppSdk. If you exclude WinAppSdk but have Gtk & Linux FB they will duplicate the files rather than using the same version between projects.

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
dansiegel: this was some debugging code... thanks for catching it

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp/MainPage.xaml
```diff
@@ -0,0 +1,14 @@
+<Page x:Class="MyExtensionsApp.MainPage"
+	xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+	xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+	xmlns:local="using:MyExtensionsApp"
+	xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+	xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+	mc:Ignorable="d"
+	Background="{ThemeResource BackgroundBrush}">
```
nickrandolph, line 8: BackgroundBrush doesn't exist so will cause WinUI to exception out with parse error. Uno seems more forgiving.
Use ApplicationPageBackgroundThemeBrush instead

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Windows/Properties/launchSettings.json
```diff
@@ -1,7 +0,0 @@
-{
-  "profiles": {
-	"MyExtensionsApp.Windows (Package)": {
```
nickrandolph, line 3: Why has this file and the pubxml files been removed. Please add them back

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Windows/MyExtensionsApp.Windows.csproj
```diff
@@ -3,97 +3,102 @@
...
-	</PropertyGroup>
-
-	<PropertyGroup>
-		<!-- Bundles the WinAppSDK binaries (Uncomment for unpackaged builds) -->
-		<!-- <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained> -->
-		<!-- This bundles the .NET Core libraries (Uncomment for packaged builds)  -->
-		<SelfContained>true</SelfContained>
-		<DefaultLanguage>en</DefaultLanguage>
+		<WindowsPackageType>MSIX</WindowsPackageType>
+		<WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>
```
nickrandolph, line 13: I know this line (WindowsAppSDKSelfContained) is in the unoapp template, but please remove from here. I don't believe we should be encouraging this as best practice - for store submission apps should have this as false to minimize download size.

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.xaml
```diff
@@ -16,19 +16,27 @@
 			<ResourceDictionary.MergedDictionaries>
 				<!-- Load WinUI resources -->
 				<XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
-
+<!--#if (useDefaultAppTemplate)-->
+<!--#if (useMaterial)-->
 				<MaterialColors xmlns="using:Uno.Material"
-								OverrideSource="ms-appx:///Styles/ColorPaletteOverride.xaml" />
+								OverrideSource="ms-appx:///MyExtensionsApp/Styles/ColorPaletteOverride.xaml" />
```
nickrandolph, line 22: OverrideSource is causing issues with Windows. For the moment, just comment out the overridesource eg

				<MaterialColors xmlns="using:Uno.Material" />
				
				<MaterialFonts xmlns="using:Uno.Material" />

nickrandolph (approved):

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Windows/Properties/launchSettings.json
dansiegel: they got git ignored when there were moved.

PR #1039, opened 2023-01-05
fix: Using StorageHelpers to determine if file exists in uno app package

nickrandolph:
GitHub Issue (If applicable): #927

## PR Type

- Bugfix

Nested files packages with the application can't be accessed using IStorage

## What is the new behavior?

Nested files are accessible using IStorage

## PR Checklist

## Other information

### DIFF
--- a/samples/Playground/Directory.Packages.props
+++ b/samples/Playground/Directory.Packages.props
@@ -29,16 +29,16 @@
 		<PackageVersion Include="Uno.Toolkit.UI.Material" Version="2.4.2" />
 		<PackageVersion Include="Uno.Toolkit.WinUI" Version="2.4.2" />
 		<PackageVersion Include="Uno.Toolkit.WinUI.Material" Version="2.4.2" />
-		<PackageVersion Include="Uno.UI" Version="4.7.0-dev.918" />
-		<PackageVersion Include="Uno.UI.Adapter.Microsoft.Extensions.Logging" Version="4.7.0-dev.918" />
-		<PackageVersion Include="Uno.UI.RemoteControl" Version="4.7.0-dev.918" />
-		<PackageVersion Include="Uno.UI.Skia.Gtk" Version="4.7.0-dev.918" />
-		<PackageVersion Include="Uno.UI.Skia.Tizen" Version="4.7.0-dev.918" />
-		<PackageVersion Include="Uno.UI.Skia.Wpf" Version="4.7.0-dev.918" />
-		<PackageVersion Include="Uno.UI.WebAssembly" Version="4.7.0-dev.918" />
+		<PackageVersion Include="Uno.UI" Version="4.7.0-dev.972" />
+		<PackageVersion Include="Uno.UI.Adapter.Microsoft.Extensions.Logging" Version="4.7.0-dev.972" />
+		<PackageVersion Include="Uno.UI.RemoteControl" Version="4.7.0-dev.972" />
+		<PackageVersion Include="Uno.UI.Skia.Gtk" Version="4.7.0-dev.972" />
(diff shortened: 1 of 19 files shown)


nickrandolph:
@arqueror I've rebased this PR and applied the changes that were made. I think there's a build issue related to the generated code - will have to wait for build to finish

dr1rrb:
For me there is an issue in the code generated by uno, most probably a missing `global::` prefix, which drives to conflict with our code which is also in the `Uno` namespace. This seems to be the generator which determines which `OnXXXX` methods are overrided by the control (so we have to subscribe to the event on control's creation).

The issue is the same for the `FeedView` and the `FrameView`.

Youssef1313:
It's not quite clear to me what the problem is.

Here is a unit test showing an example of the generated code:

https://github.com/unoplatform/uno/blob/bbd92e7bc7fe70e838a4eac0148e2c1e06b8b6a2/src/SourceGenerators/Uno.UI.SourceGenerators.netcore.Tests/ImplementedRoutedEventsGeneratorTests/Given_ImplementedRoutedEventsGenerator.cs#L71-L96

nickrandolph:
@Youssef1313 it only appears to be an issue on the UWP build I think. This PR picked up the latest Uno builds from yesterday, so looks like https://github.com/unoplatform/uno/pull/10771 didn't completely fix the issue

Youssef1313:
@nickrandolph Does `<TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>` fix things when set for Windows builds?

nickrandolph:
Why do we need to set this now? We've never had to set this in the past? The break in this PR is just due to updating Uno version as far as I can tell

Youssef1313:
> Why do we need to set this now?

The root issue exists in earlier version of Uno but was quite hidden. If you look at `samples-windows.binlog` from CI artifacts, you'll see the following

image

as you can see, XamlCodeGenerator has crashed, but the crash is silent and doesn't fail the build. It shouldn't have been run in the first place.

For `ImplementedRoutedEventsGenerator`, it's run but it doesn't crash. So it generated invalid code.

So the issue isn't really recent.

Is setting the property the correct action to take? I don't know, but if it fixes the build, it could be used as a temporary workaround at least.

Youssef1313:
Alternatively:


```xml
<ItemGroup Condition="SomeConditionHereForWindowsBuilds">
  <Analyzer Remove="@(Analyzer)" Condition="'%(Analyzer.NugetPackageId)'=='Uno.UI'"/>
  <Analyzer Remove="@(Analyzer)" Condition="'%(Analyzer.NugetPackageId)'=='Uno.WinUI'"/>
</ItemGroup>
```

Ideally, I don't think Windows builds should reference Uno.UI at all. Looking at the package, it looks like Uno.UI for UAP only contains `Uno.UI.Toolkit`. So, moving `Uno.UI.Toolkit` to its own package makes the most sense to me, then in Windows builds, only Uno.UI.Toolkit is referenced. This has the advantage of not having to pull the whole Uno.UI package when restore is only done for Windows build, and also avoids including source generators for Windows builds. That is a major change that may or may not happen, and if it happened, I don't think it will be soon.

Youssef1313:
I opened https://github.com/unoplatform/uno/pull/10932 as a fix on Uno side.

nickrandolph:
@Youssef1313 I've used the uno libraries from your PR and seeing same build issue. I've sent you the binlog

Youssef1313:
@nickrandolph Consider testing with 4.7.0-dev.908

nickrandolph:
I will give it a shot later this evening

nickrandolph:
I will give it a shot later this evening

arqueror (approved):

dansiegel (approved):

PR #1046, opened 2023-01-06
feat: Removes test projects and data contract project from blank template

### LINKED ISSUE #1036: [Template] Tidy up "blank" template
The blank template shouldn't include DataContract or Test projects
image
(since the WeatherForecast class is used by the WeatherController, we'll need to include that in the server project in this scenario)

dansiegel:

- closes #1036

## PR Type

- Refactoring

DataContracts, Unit Test and UI Tests get included with blank template

## What is the new behavior?

- DataContracts project is excluded from the solution
- WeatherForecast model is part of the Server project when using Blank Template Type
- Serilog is disabled on the server when using a Blank Template Type
- Unit Tests project is excluded from the solution
- UI Tests project is excluded from the solution

### DIFF
--- a/src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
+++ b/src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
@@ -377,7 +377,7 @@
     "useSerilog": {
       "type": "computed",
       "datatype": "bool",
-      "value": "(logging && serilog)"
+      "value": "(logging && serilog && templateType == 'default')"
     },
     "useMvux": {
       "type": "computed",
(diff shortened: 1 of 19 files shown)


src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -377,7 +377,7 @@
     "useSerilog": {
       "type": "computed",
       "datatype": "bool",
-      "value": "(logging && serilog)"
+      "value": "(logging && serilog && templateType == 'default')"
```
nickrandolph, line 380: Should this be "useDefaultAppTemplate" instead of templateType=='default'

nickrandolph (approved):

PR #1068, opened 2023-01-13
Migrate HostBuilder code to AppStart

### LINKED ISSUE #1063: [Template] Move the host builder code out of App.xaml.cs into file in the class library

Move host builder code into a file in the class library project

## Why is this needed:

Currently devs have to go into the target projects in order to access app.xaml.cs in order to modify host builder.

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

dansiegel:

- closes #1063

## PR Type

- Refactoring

App HostBuilder code is done inside of the `App.OnLaunched`. Because this is linked to Platform heads it requires opening a platform project to edit any initialization code that is part of the host builder / initial navigation.

## What is the new behavior?

When using the default template this code has been migrated to a new AppStart class in the multi-targeted project where the main app code lives. This makes it easier for developers to stay within the context of the single project to edit anything they need.


src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Base/App.default.cs
```diff
@@ -19,170 +19,11 @@ public App()
...
-			// Ensure the current window is active
-			_window.Activate();
-		}
-//+:cnd:noEmit
-#else
-		_host = await builder.NavigateAsync<Shell>();
-#endif
-	}
-
 #if useFrameNav
```
nickrandolph, line 22: Do we still need this conditional logic?

nickrandolph (approved):

PR #1085, opened 2023-01-18
Add missing launchSettings.json for WASM template

dansiegel:

n/a

## PR Type

- Bugfix

launchSettings.json is missing from WASM target

## What is the new behavior?

launchSettings.json has been added for WASM target and inspect uri has been added for Server when it is hosting the WASM head

### DIFF
--- a/src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Wasm/Properties/launchSettings.json
+++ b/src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Wasm/Properties/launchSettings.json
@@ -1,25 +1,19 @@
 {
-  "$schema": "https://json.schemastore.org/launchsettings.json",
   "iisSettings": {
     "windowsAuthentication": false,
     "anonymousAuthentication": true,
     "iisExpress": {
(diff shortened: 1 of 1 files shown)


src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Server/Properties/launchSettings.json
```diff
@@ -28,6 +29,7 @@
       "launchBrowser": true,
       //#if wasm
       "launchUrl": "",
```
nickrandolph, line 31: Is the url for wasm configurable... I'm wondering whether the assumption that the wasm app will be at the root is reasonable or not.

nickrandolph (approved):

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Server/Properties/launchSettings.json
dansiegel: out of scope.... as we discussed we can create an issue and provide some guidance in the template for how to change the root of the hosted WASM app if someone so desires.

PR #1092, opened 2023-01-19
Dev/ds/parameter consolidation

### LINKED ISSUE #1042: [Template] Include mvu-x without Hosting and other extensions

Would like the ability to include options such as mvu-x without having Hosting and other extensions included.

Currently we have the blank template type (which doesn't include any of the extensions) or we have default/recommended (which includes hosting at a minimum).

This should expose a "Hosting" option which would control whether the template includes host builder. Most other extensions would require this option to be set to true in order for them to be supported

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

### LINKED ISSUE #1043: [Template] Refactor symbol usage in template code

Refactor template code to only reference computed symbols.

## Why is this needed:

Currently the template code uses a mix of input symbols (eg configuration) and computed symbols (eg useConfiguration). If we change the input symbols (eg renaming) the template code has to be updated. By only using computed symbols in the template code we're able to change the input symbols without adjusting the template code

## For which Platform:

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Build tasks

### LINKED ISSUE #1090: [Template] Link IsExternalInit in template

Move the IsExternalInit to the Base directory and link to it in the DataContracts and Multi-Targeted project.

## Why is this needed:

It currently only exists in DataContracts as it is needed for the template to work but would also be helpful for the Multi-targeted project as we also use netstandard as a target there and it makes it more seamless to get started.

dansiegel:

- closes #1090 
- closes #1043
- closes #1042

## PR Type

- Refactoring
- Feature

Complex logical checks are needed throughout the template and Dependency Injection is required with the default template.

## What is the new behavior?

Additional Symbols have been added to validation whether or not features should be turned on or off. Root dependencies turn off dependent features. For example disabling Configuration will disable Http & Localization extensions. An additional parameter for Dependency Injection has been added. When disabled


src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -362,22 +388,27 @@
...
-      "value": "(appTheme == 'fluent')"
+      "value": "(appTheme == 'fluent' && templateType == 'default')"
     },
     "useCupertino": {
       "type": "computed",
       "datatype": "bool",
-      "value": "(appTheme == 'cupertino')"
+      "value": "(appTheme == 'cupertino' && templateType == 'default')"
+    },
+    "useLogging": {
```
nickrandolph, line 403: As a fall back for when useDependencyInjection is false, we could use the logging block that's in the unoapp template

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.Mobile/MyExtensionsApp.Mobile.csproj
```diff
@@ -1,43 +1,10 @@
...
-		<!--#else-->
-		<!--#if (android)-->
-		<TargetFrameworks>$(TargetFrameworks);net6.0-android</TargetFrameworks>
-		<!--#endif-->
-		<!--#if (ios)-->
-		<TargetFrameworks>$(TargetFrameworks);net6.0-ios</TargetFrameworks>
-		<!--#endif-->
+		<TargetFrameworks>$mobileTargetFrameworks$</TargetFrameworks>
 		<!--#if (macos)-->
 		<!-- Disabled because of https://github.com/xamarin/xamarin-macios/issues/16401-->
```
nickrandolph, line 5: Need to double-check with Jerome that this is still the case after recent update came out

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp/MyExtensionsApp.csproj
```diff
@@ -125,33 +134,37 @@
...
 		<Compile Update="**\*.xaml.cs">
 			<DependentUpon>%(Filename)</DependentUpon>
 		</Compile>
 	</ItemGroup>
+	<!--#endif-->
 
 	<ItemGroup>
-		<!--#if (useDefaultAppTemplate)-->
 		<!--#if (useConfiguration)-->
 		<None Remove="appsettings*.json" />
```
nickrandolph, line 163: I think we can remove the Remove from None

src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp/MyExtensionsApp.csproj
```diff
@@ -125,33 +134,37 @@
...
 		<EmbeddedResource Include="appsettings.json" />
 		<EmbeddedResource Include="appsettings.*.json"
 											DependentUpon="appsettings.json" />
 		<!--#endif-->
-		<!--#endif-->
 		<PRIResource Include="Strings\**\*.resw" />
```
nickrandolph, line 168: Is this still required - check latest unoapp template

nickrandolph (approved):

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
dansiegel: can you create an issue for that and we can handle the logging change in a new PR

PR #1121, opened 2023-01-29
[Template] Net 7 updates

dansiegel:
## PR Type

- Bugfix

The template sets the LangVersion to 10 regardless of whether you are targeting net6 or net7. We need to set this due to the use of netstandard when using net6 but for the net7 version it is not needed as net7 uses 11 as a default. We also specify 4.2 for the Compiler Toolset.

## What is the new behavior?

We use the default Compiler Toolset and only specify the LangVersion for net6.0


src/Uno.Extensions.Templates/content/unoapp-extensions/Directory.Build.props
```diff
@@ -41,9 +45,16 @@
 		<SupportedOSPlatformVersion Condition="$(IsWinAppSdk)">10.0.18362.0</SupportedOSPlatformVersion>
 		<TargetPlatformMinVersion Condition="$(IsWinAppSdk)">10.0.18362.0</TargetPlatformMinVersion>
 	</PropertyGroup>
+	<!--#if (useNetSix)-->
 
 	<ItemGroup>
 		<!--This override is used to validate the use of specific version of the C# Compiler-->
+		<!--#if (cpm)-->
```
nickrandolph, line 52: Should this be a computed value eg useCPM instead of cpm?

dansiegel: if it was computed it would simply equal cpm.. not sure what value we would get from it.

nickrandolph (approved):

PR #1187, opened 2023-02-21
feat: Updating text for template

nickrandolph:
Review of text for template

### DIFF
--- a/src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
+++ b/src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
@@ -415,7 +415,7 @@
       "type": "parameter",
       "dataType": "bool",
       "defaultValue": "true",
-      "description": "Adds the Visual Studio Code Debugging support files for WebAssembly and Skia"
+      "description": "Adds the Visual Studio Code Debugging support files for WebAssembly and Skia Gtk"
     },
     "wasmPwaManifest": {
       "displayName": "PWA Manifest",
--- a/src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
+++ b/src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
@@ -57,22 +57,22 @@
       "choices": [
         {
           "choice": "recommended",
-          "description": "A basic project using Uno.Extensions",
+          "description": "An extended template for creating a multi-platform application with the Uno Platform",
           "displayName": "Default"
         },
         {
           "choice": "blank",
-          "description": "A Blank app without Uno.Extensions",
+          "description": "A template for creating a multi-platform application with the Uno Platform",
           "displayName": "Blank"
         }
       ]
     },
     "appId": {
-      "displayName": "App ID",
+      "displayName": "Application ID",
       "type": "parameter",
       "datatype": "string",
       "replaces": "com.companyname.myextensionsapp",
-      "description": "The application identifier (e.g. com.mycompany.myapp).",
+      "description": "The application identifier (e.g. com.mycompany.myapp)",
       "defaultValue": "com.companyname.myextensionsapp"
     },
     "tfm": {
@@ -86,12 +86,12 @@
       "choices": [
         {
           "choice": "net6.0",
-          "displayName": ".NET 6.0 (Long Term Support)",
+          "displayName": ".NET 6.0",
           "description": "Target .NET 6.0 (Long Term Support)"
         },
         {
           "choice": "net7.0",
-          "displayName": ".NET 7.0 (Standard Term Support)",
+          "displayName": ".NET 7.0",
           "description": "Target .NET 7.0 (Standard Term Support)"
         }
       ]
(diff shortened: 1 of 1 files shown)


src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -86,12 +86,12 @@
       "choices": [
         {
           "choice": "net6.0",
-          "displayName": ".NET 6.0 (Long Term Support)",
+          "displayName": ".NET 6.0",
           "description": "Target .NET 6.0 (Long Term Support)"
         },
         {
           "choice": "net7.0",
-          "displayName": ".NET 7.0 (Standard Term Support)",
```
dansiegel, line 94: These values were specifically chosen to mimic the values you would get from an Official Microsoft template
image

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -417,18 +408,18 @@
...
+      "description": "Use Central Package Management where package versions are via central Directory.Packages.props file"
     },
     "vscode": {
-      "displayName": "WebAssembly - Visual Studio Code Debugging",
+      "displayName": "Visual Studio Code Debugging",
       "type": "parameter",
       "dataType": "bool",
       "defaultValue": "true",
-      "description": "Adds the Visual Studio Code Debugging support files for WebAssembly"
+      "description": "Adds the Visual Studio Code Debugging support files for WebAssembly and Skia"
```
dansiegel, line 418: As the tasks.json and launch.json only have something for WASM and Skia GTK, we should either:
1. Change this to specify Skia Gtk instead of just 'Skia'
2. Update the .vscode files to "support" the other Skia targets

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -417,18 +408,18 @@
...
     },
     "wasmPwaManifest": {
-      "displayName": "WebAssembly - PWA Manifest",
-      "description": "Enables the generation of a PWA manifest for WebAssembly",
+      "displayName": "PWA Manifest",
+      "description": "Generates a PWA manifest for WebAssembly",
```
dansiegel, line 422: Probably better to mimic the terminology we have for the Test projects `Include` instead of `Generates`

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -370,44 +361,44 @@
...
-          "displayName": "None"
+          "description": "Logging is configured for the application",
+          "displayName": "Basic"
         },
         {
           "choice": "default",
-          "description": "Enables basic logging to the console",
-          "displayName": "Default"
+          "description": "Logging to the console is configured",
+          "displayName": "Console"
```
dansiegel, line 378: Reading these 2 descriptions I would be very confused as to what these mean/do.. and I'm the one that set it up 🤣 
Remember technically they will get the same Console logging either way. The default actually wires things up with DI to use throughout their app where the Core / Basic is getting some logging output from the framework.

dansiegel (changes requested):

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
nickrandolph: We're not using that UI, so irrelevant - it's still included in the description for that reason

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
nickrandolph: Yeh I was struggling to a) give these options a name and b) a sensible description

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
nickrandolph: Updated - good suggestion

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
nickrandolph: Updated but I'm sure it could be reworked more

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
nickrandolph: Updated to Skia Gtk

src/Uno.Extensions.Templates/content/unoapp-extensions/.template.config/template.json
```diff
@@ -370,44 +361,44 @@
...
+      "description": "Configures logging support in the application",
       "type": "parameter",
       "datatype": "choice",
       "defaultValue": "serilog",
       "choices": [
         {
           "choice": "none",
-          "description": "Disables logging",
-          "displayName": "None"
+          "description": "Logging writes output to Console",
```
dansiegel, line 372: Much better... What do you think about this though. The logging that gets enabled in this scenario is only included for Debug builds and it's added specifically to the App Framework... so probably not something that you would use in your app for logging in general.
```suggestion
          "description": "Debug Logging for the Application Framework to the Console",
```

dansiegel (approved):

PR #1193, opened 2023-02-23
fix(gen): Better incrementality

Youssef1313:
Not yet completely good. But a step forward.

## PR Type

## PR Checklist

## Other information


src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -12,32 +14,78 @@ namespace Uno.Extensions.Generators.PropertySelector;
...
 
-	public GeneratorSyntaxContext Context { get; }
-
 	public InvocationExpressionSyntax? Syntax { get; }
 	public FileLinePositionSpan Location { get; }
-
-	public IMethodSymbol? Method { get; }
+	public string? MethodGlobalNamespace { get; }
+	public string? MethodName { get; }
+	public ImmutableList<PropertySelectorUsageItem>? Selectors { get; }
```
Youssef1313, line 90: Note to self so that I don't forget. `ImmutableList` is not the correct type to use here. We need a custom equatable data structure.

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -1,43 +1,207 @@
...
 using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Uno.Extensions.Core.Generators.Helpers;
+using Uno.Extensions.Edition;
 
 namespace Uno.Extensions.Generators.PropertySelector;
 
 internal readonly record struct PropertySelectorCandidate
```
Youssef1313, line 16: This model has more fields now, and I'm not very sure if `struct` is still good here. @dr1rrb should we switch this to a record class?

Side note: One issue with this generator is that it is triggered by `InvocationExpressionSyntax`, and we have no way to filter them down to few ones syntactically. So in any case, we have to create the model for **every** invocation in the compilation and get semantic info about the model being called for **every** invocation.

The good thing about this generator though is that it uses `RegisterImplementationSourceOutput`, so a good host will not re-run on every typed character. So this is effectively run only on build (can run in niche scenarios like EnC/HotReload when the IDE cares about the code being generated).

dansiegel (commented):
Be sure to rebase this

dansiegel (changes requested):

src/Uno.Extensions.Core.Generators/Analyzers/PropertySelectorCanOnlyUsePropertyMembersAnalyzer.cs
```diff
@@ -0,0 +1,216 @@
...
+		public override void VisitMemberBindingExpression(MemberBindingExpressionSyntax node)
+		{
+		}
+
+		public override void VisitPostfixUnaryExpression(PostfixUnaryExpressionSyntax node)
+		{
+			Visit(node.Operand);
+			if (!node.IsKind(SyntaxKind.SuppressNullableWarningExpression))
+			{
+				Debugger.Launch();
```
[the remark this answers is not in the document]
Youssef1313: I didn't intend to commit it 😄 

```suggestion
```

nickrandolph:
@dr1rrb if you have a minute can you review this PR and confirm that it's ok to be merged (or advise on what changes might be required)?

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -1,43 +1,201 @@
...
+
+		static (bool hasDefault, string? value) GetDefault(IParameterSymbol param)
+			=> param switch
+			{
+				{ HasExplicitDefaultValue: true } => (true, null),
+				{ Type.NullableAnnotation: NullableAnnotation.Annotated } => (true, $"default({param.Type})"),
+				{ } p when p.Type.IsNullable() => (true, $"default({param.Type})"),
+				_ => (false, null)
+			};
+	}
```
dr1rrb, line 200: I would move the generation code itself into a dedicated helper class. I would not expect to have such code in a `Candidate` class (e.g. keep it the `PropertySelectorsGenerationTool` but making it internal static)

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -1,43 +1,201 @@
...
+			return $@"(current_0, updated_{count}) =>
+			{{
+				{current.Align(4)}
+				{updated.Align(4)}
+
+				return updated_0;
+			}}".Align(0);
+		}
+
+		static string OrDefault(ITypeSymbol type)
```
dr1rrb, line 163:
```suggestion
		private static string OrDefault(ITypeSymbol type)
```

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -1,43 +1,201 @@
...
+			.Where(param => param.Type is INamedTypeSymbol
+			{
+				IsGenericType: true,
+				MetadataName: "PropertySelector`2",
+				ContainingNamespace: { Name: "Edition", ContainingNamespace: { Name: "Extensions", ContainingNamespace.Name: "Uno" } }
+			})
+			.Select(param =>
+				(key: $"{param.Name}{callerPath}{callerLine}", accessor: GenerateAccessor((INamedTypeSymbol)param.Type, arguments.ElementAtOrDefault(param.Ordinal))))
+			.Where(a => a.accessor is not null)
+			.ToImmutableArray().AsEquatableArray()!;
```
dr1rrb, line 73:
```suggestion
			.ToImmutableArray()
			.AsEquatableArray()!;
```

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -1,43 +1,201 @@
...
-	public GeneratorSyntaxContext Context { get; }
+		Accessors = method.Parameters
+			.Where(param => param.Type is INamedTypeSymbol
+			{
+				IsGenericType: true,
+				MetadataName: "PropertySelector`2",
+				ContainingNamespace: { Name: "Edition", ContainingNamespace: { Name: "Extensions", ContainingNamespace.Name: "Uno" } }
+			})
+			.Select(param =>
+				(key: $"{param.Name}{callerPath}{callerLine}", accessor: GenerateAccessor((INamedTypeSymbol)param.Type, arguments.ElementAtOrDefault(param.Ordinal))))
```
dr1rrb, line 71: Could / should we defer the generation of the code until really needed?

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
dr1rrb: For the `class` vs `struct`, your call :) I may be wrong, but I've the feeling that some of those properties are no longer used since you generate the code much earlier.

src/Uno.Extensions.Core.Generators/PropertySelector/Rules.PropertySelector.cs
```diff
@@ -193,18 +173,18 @@ public static class PS0102
...
 			=> Diagnostic.Create(
 				Descriptor,
 				Location.Create(argNode.SyntaxTree, argNode.Span),
 				param,
-				method.ToString());
+				method);
 
-		public static GenerationException FailFileArg(IMethodSymbol method, IParameterSymbol? param, SyntaxNode argNode)
-			=> new(GetDiagnostic(method, "[CallerFilePath] " + (param?.Name ?? ""), argNode));
+		public static Diagnostic FailFileArg(IMethodSymbol method, IParameterSymbol? param, SyntaxNode argNode)
```
dr1rrb, line 183:
```suggestion
		public static Diagnostic GetFileArgDiagnostic(IMethodSymbol method, IParameterSymbol? param, SyntaxNode argNode)
```

As you removed the `Fail` methods we can consider to renamed all `GetDiagnostic` to have something shorter, but we need to stay consistent :)

src/Uno.Extensions.Core.Generators/PropertySelector/Rules.PropertySelector.cs
```diff
@@ -193,18 +173,18 @@ public static class PS0102
...
+			=> GetDiagnostic(method.ToString(), "[CallerFilePath] " + (param?.Name ?? ""), argNode);
 
-		public static GenerationException FailLineArg(IMethodSymbol method, IParameterSymbol? param, SyntaxNode argNode)
-			=> new(GetDiagnostic(method, "[CallerLineNumber] " + (param?.Name ?? ""), argNode));
+		public static Diagnostic FailLineArg(IMethodSymbol method, IParameterSymbol? param, SyntaxNode argNode)
```
dr1rrb, line 186:
```suggestion
		public static Diagnostic GetLineArgDiagnostic(IMethodSymbol method, IParameterSymbol? param, SyntaxNode argNode)
```

dr1rrb (approved):

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -1,43 +1,201 @@
...
+
+			if (ctor is null)
+			{
+				return "";
+			}
+
+			return $" ?? new({ctor.Where(arg => arg.value is not null).Select(arg => arg.value).JoinBy(", ")})";
+		}
+
+		static (bool hasDefault, string? value) GetDefault(IParameterSymbol param)
```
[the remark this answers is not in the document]
Youssef1313: @dr1rrb It is a local function. Do you want to move it out?

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
Youssef1313: @dr1rrb It is a local function. Do you want to move it out?

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
Youssef1313: @dr1rrb I guess it will always be needed. When we reach here, that means it is a valid candidate, which means we will be calling `Generate` which will need `Accessors`

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
Youssef1313: SyntaxNode wasn't used indeed. But the rest appears to be used.

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
```diff
@@ -1,43 +1,201 @@
...
+		if (method is null ||
+			!method.Parameters.Any(p => p.Type is INamedTypeSymbol
+			{
+				IsGenericType: true,
+				MetadataName: "PropertySelector`2",
+				ContainingNamespace: { Name: "Edition", ContainingNamespace: { Name: "Extensions", ContainingNamespace.Name: "Uno" } }
+			}) ||
+			method.Parameters.FirstOrDefault(p => p.HasAttribute<CallerFilePathAttribute>()) is not { } callerPathParameter ||
+			method.Parameters.FirstOrDefault(p => p.HasAttribute<CallerLineNumberAttribute>()) is not { } callerLineParameter
+			)
```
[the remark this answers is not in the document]
Youssef1313: Logic-wise they are close, but analyzer implementation relies more on symbols.

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorsGenerationTool.cs
```diff
@@ -14,49 +13,35 @@ internal class PropertySelectorsGenerationTool : ICodeGenTool
 	/// <inheritdoc />
 	public string Version => "1";
```
Youssef1313, line 14: Not sure if I should increment this?

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
dr1rrb: Oh my bad, I didn't realized that. No need to move it out ^^

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorsGenerationTool.cs
dr1rrb: We could, but it's only for the `[GeneratedCode]` and other generated headers ... I don't think it has a significant impact

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorCandidate.cs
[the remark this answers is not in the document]
dr1rrb: Then could we extract them in an helper? The idea I that I've the feeling we could easily update one and forget to update the other one as they are in 2 different classes and not even sharing the same name.

But I may be wrong: the 2 classes are searching for the same nodes but in a different way. I didn't validated but maybe there are plenty of other places in those classes that should be shared to avoid such issue, so it's irrelevant to do it only for this part? Your call :)

src/Uno.Extensions.Core.Generators/PropertySelector/PropertySelectorsGenerationTool.cs
```diff
@@ -234,7 +193,7 @@ static string OrDefault(PropertySelectorPath path, PropertySelectorPathPart part
 
 			if (ctor is null)
 			{
-				throw Rules.PS0006.Fail(path.FullPath, part.Name, part.Node, type);
+				return "";
```
dr1rrb, line 196:
```suggestion
				return ""; // PS006 Failed
```

dr1rrb (approved):
LGTM, @Youssef1313 you should address remaining comments and we can merge it now that v2.3 is out now.

Youssef1313:
@dr1rrb @dansiegel Is this ready to merge?

dansiegel (approved):

PR #1234, opened 2023-03-06
feat: adding serialization example

dansiegel:

- #1113

## PR Type

- Feature

There is no use of the Serialization API from Uno.Extensions in the template

## What is the new behavior?

When using the Http Serialization gets wired up with the JsonSerializerContext. This is also enabled for the Server. An IWeatherCache has been added using the ISerializer and pulls from the latest cache if the app is offline.


src/Uno.Extensions.Templates/content/unoapp-extensions/MyExtensionsApp.DataContracts/WeatherForecast.cs
```diff
@@ -1,5 +1,33 @@
...
+
+	/// <summary>
+	/// Get a description of how the weather will feel.
+	/// </summary>
+	public string? Summary { get; set; }
+
+	/// <summary>
+	/// Gets the Forecast Temperature in Fahrenheit
+	/// </summary>
+	public int TemperatureF => 32 + (int)(TemperatureC / 0.5556);
```
nickrandolph, line 28: This still annoys me....

PR #1265, opened 2023-03-13
[Template] Minimal API

### LINKED ISSUE #1264: [Template] Use AspNetCore Minimal Apis
Add option for AspNetCore Minimal Apis

## Why is this needed:

These are the standard default for new AspNetCore API projects from the official Microsoft template. We should add this support for the Server Project in the Extensions Template

dansiegel:

- closes #1264

## PR Type

- Feature

The server project provides a standard Controller for the WeatherForecast API

## What is the new behavior?

By default you now get a Minimal Api with the same Open Api support you get from the existing Controller. You can additionally disable minimal API's to get the existing controller. This also updates the WeatherForecast model to use DateOnly for .NET 7.0+. We cannot use DateOnly for .NET 6.0 due to the library having to use netstandard2.1 instead of being able to use net6.0.

### DIFF
--- a/build/ci/.azure-pipelines.TemplateValidation.yml
+++ b/build/ci/.azure-pipelines.TemplateValidation.yml
@@ -55,6 +55,12 @@ jobs:
       FrameNavigation:
         createdProjectName: UnoAppFrameNavigation01
         templateArgs: '--navigation blank'
+      ServerController:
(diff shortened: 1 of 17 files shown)


jeromelaban (approved):
👍

PR #1321, opened 2023-03-23
docs: Fix class name

dr1rrb:
docs: Fix class name


lukeblevins (approved):
Thanks to @jeanplevesque for spotting this

Issue #1469, opened 2023-05-17
[Navigation] x:Bind doesn't work for Page


nickrandolph:
## Current behavior

Attempt to x:Bind to a property on a viewmodel associated with a page - this fails to set the value in some cases because the DataContext is sometimes null when bindings are initialized.

## Expected behavior

x:Bind should work

## How to reproduce it (as minimally and precisely as possible)

Create new app using wizard
Pick default option and customize:
- Mvvm
- Remove Server and Http
Add ViewModel property to MainPage.xaml.cs:
`public MainViewModel? ViewModel => DataContext as MainViewModel;`
Change Binding to x:Bind in MainPage.xaml
`<utu:NavigationBar Content="{x:Bind ViewModel.Title}" />`
Add delay to MainViewModel constructor

```
        public MainViewModel(....)
        {
            ... 
            for (int i = 0; i < 2000; i++)
            {
                System.Diagnostics.Debug.WriteLine(i);
            }
        }
```

No NavigationBar Title:
![image](https://github.com/unoplatform/uno.extensions/assets/1614057/60bbd87a-8eeb-4d17-9acd-b27aac92af02)

## Environment

Nuget Package:

Package Version(s):

Affected platform(s):

- [ ] iOS
- [ ] Android
- [ ] WebAssembly
- [ ] WebAssembly renders for Xamarin.Forms
- [ ] Windows
- [ ] Build tasks

Visual Studio:

- [ ] 2017 (version: )
- [ ] 2019 (version: )
- [ ] for Mac (version: )

Relevant plugins:

- [ ] Resharper (version: )

## Anything else we need to know?

Workaround - Update MainPage.xaml.cs to call Bindings.Update when DataContext changes

```
public MainPage()
{
    this.InitializeComponent();

    DataContextChanged += (_, _) => Bindings.Update();
}
```

jeromelaban:
This is expected, and related to https://github.com/unoplatform/uno/issues/2872 and https://github.com/unoplatform/uno/issues/2895

Youssef1313:
@jeromelaban It happens on Windows too.

nickrandolph:
> This is expected, and related to [unoplatform/uno#2872](https://github.com/unoplatform/uno/issues/2872) and [unoplatform/uno#2895](https://github.com/unoplatform/uno/issues/2895)

[...]

PR #1478, opened 2023-05-19
fix: Adding IForceBindingUpdate code generation to fix issue with x:Bind in Navigation

nickrandolph:
GitHub Issue (If applicable): #1469 

## PR Type

- Bugfix

With Navigation x:Bind if constructor of ViewModel takes too long

## What is the new behavior?

Bindings.Update is called after viewmodel set to force update of x:Bind bindings

## PR Checklist

## Other information


src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+
+	private static string GetViewModelName(INamedTypeSymbol type)
+		=> $"{type.Name}";
+
+	private string Generate(INamedTypeSymbol model)
+	{
+		var className = GetViewModelName(model);
+
```
Youssef1313, line 48:
```suggestion
	private string Generate(INamedTypeSymbol model)
	{
		var className = model.Name;
```

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -0,0 +1,66 @@
...
+		var updateInterface = _ctx.ForceBindingsUpdateInterface.ToDisplayString();
+		var fileCode = this.AsPartialOf(
+			model,
+			attributes: default,
+			bases: updateInterface,
+			code: $@"
+				async ValueTask IForceBindingsUpdate.ForceBindingsUpdateAsync()
+				{{	
+					this.Bindings.Update();
+				}}
```
Youssef1313, line 59: Why is this async?

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,71 @@
...
+	private static Regex ClassRegEx = new Regex("x:Class=\"([\\w.]+)\"");
+	private static Regex xBindRegEx = new Regex("{x:Bind ");
+
+	private IDictionary<string, AdditionalText> XBindFiles
+	{
+		get
+		{
+			if (_XBindFiles is null)
+			{
+				var files = new Dictionary<string, AdditionalText>();
```
Youssef1313, line 39: Consider using `ImmutableDictionary.CreateBuilder` rather than a Dictionary.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,71 @@
...
+			return _XBindFiles;
+		}
+	}
+
+	public bool ContainsXBind(INamedTypeSymbol symbol)
+	{
+		try
+		{
+			var name = symbol.ToDisplayString();
+			return XBindFiles.ContainsKey(name);
```
Youssef1313, line 65: I think only the value (AdditionalText) from the dictionary is never accessed? Consider switching to a hash set

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,71 @@
...
+		}
+		catch { }
```
Youssef1313, line 67: I can't see any exception the above code can throw. `ContainsKey` throws when the argument is null, but `name` shouldn't be null.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerator.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.CodeAnalysis;
+using Uno.Extensions.Generators;
+
+namespace Uno.Extensions.Navigation.Generators;
+
+/// <summary>
+/// A generator that generates bindable VM for the reactive framework
+/// </summary>
+[Generator]
+public partial class ForceBindingsUpdateGenerator : ISourceGenerator
```
Youssef1313, line 11: If it's not much work, I think it makes more sense to implement this as an IIncrementalGenerator. It would be more difficult though, and in this case, caching should be properly tested.

src/Uno.Extensions.Navigation.UI/Controls/ForceUpdateAttribute.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Uno.Extensions.Navigation.UI.Controls;
+
+/// <summary>
+/// Flags the default constructor to use to create an instance of a record that is being de-normalized for bindings.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class)]
+public class ForceUpdateAttribute : Attribute
```
Youssef1313, line 7:
```suggestion
public sealed class ForceUpdateAttribute : Attribute
```

src/Uno.Extensions.Navigation.UI/Controls/ForceUpdateAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+	public bool IsEnabled { get; set; } = true;
+
+	/// <summary>
+	/// Configure generation the force update method.
+	/// </summary>
+	/// <param name="isEnabled">Indicates if the method should be generated or not.</param>
+	public ForceUpdateAttribute(bool isEnabled = true)
+	{
+		IsEnabled = isEnabled;
+	}
```
Youssef1313, line 21: I wouldn't introduce the same thing both as constructor argument and named argument. Only one should be needed, and actually having both is confusing IMO.

src/Uno.Extensions.Navigation/build/Package.targets
```diff
@@ -1,5 +1,8 @@
...
-    <Using Include="Uno.Extensions.Navigation" />
-  </ItemGroup>
-</Project>
\ No newline at end of file
+	<ItemGroup>
+		<AdditionalFiles Include="@(Page)" />
+	</ItemGroup>
+	<ItemGroup Condition=" '$(ImplicitUsings)' == 'true' OR '$(ImplicitUsings)' == 'enable' ">
+    	<Using Include="Uno.Extensions.Navigation" />
+	</ItemGroup>
```
Youssef1313, line 7: nit: Indentation broke here

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
nickrandolph: To make sure that if there's other implementations that are asynchronous, we don't need to change the interface in the future

Youssef1313: @nickrandolph We can keep the return value as ValueTask but make it non-async by using `ValueTask.FromResult(this.Bindings.Update())`

If other implementations wanted to async, they could. But this implementation shouldn't.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
nickrandolph: removed catch

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
nickrandolph: done

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerator.cs
nickrandolph: can look at this in future if needed

src/Uno.Extensions.Navigation.UI/Controls/ForceUpdateAttribute.cs
nickrandolph: yep, will make property read only

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -0,0 +1,62 @@
...
```
Youssef1313, line 55: Other implementations can be async, if necessary.

```suggestion
				ValueTask IForceBindingsUpdate.ForceBindingsUpdateAsync()
				{{	
					this.Bindings.Update();
					return ValueTask.CompletedTask;
				}}
```

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
nickrandolph: switched to set

src/Uno.Extensions.Navigation/build/Package.targets
nickrandolph: fixed

src/Uno.Extensions.Authentication/buildTransitive/Package.targets
Youssef1313, line 1: How does this work with the rename?

Previously, this was already packed in buildTransitive here:

https://github.com/unoplatform/uno.extensions/blob/9be16a0db11ebeff7a725553103324c80783abfa/src/Directory.Build.props#L89

But now, how it will be packed?

nickrandolph: nuget build automatically renames build\packages.target to buildTransitive\$(packageId).targets

Youssef1313, line 1: image

Unless I'm missing something, it's no longer in the package. I think the change should be reverted for now, and we might follow up with another PR to clean up how we pack props/targets

nickrandolph: Thanks for picking this up - yes, this shouldn't have been moved

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -0,0 +1,69 @@
...
+	{
+		_ctx = ctx;
+		_assembly = ctx.Context.Compilation.Assembly;
+	}
+
+	private bool IsSupported(INamedTypeSymbol? type)
+		=> type is not null
+			&& (_ctx.IsGenerationEnabled(type)
+				?? (
+						type.Is(_ctx.Page) &&
```
Youssef1313, line 27: Is the issue only for `Page`s? What about `UserControl`s?

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -0,0 +1,69 @@
...
+						type.IsPartial()
+					))
+		&& _ctx.ContainsXBind(type);
+
+	public IEnumerable<(string fileName, string code)> Generate()
+	{
+		var viewModels = from module in _assembly.Modules
+			from type in module.GetNamespaceTypes()
+			where IsSupported(type)
+			select type;
```
Youssef1313, line 37: Aren't these pages, not viewModels?

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -0,0 +1,69 @@
...
+
+		foreach (var vm in viewModels)
+		{
+			if (vm is not null)
+			{
+				yield return (vm.ToString(), Generate(vm));
```
Youssef1313, line 43: It *should* be the same, but clearer IMO.

```suggestion
				yield return (vm.ToDisplayString(), Generate(vm));
```

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,67 @@
...
+	private static Regex ClassRegEx = new Regex("x:Class=\"([\\w.]+)\"");
+	private static Regex xBindRegEx = new Regex("{x:Bind ");
+
+	private IImmutableSet<string> XBindFiles
+	{
+		get
+		{
+			if (_XBindFiles is null)
+			{
+				var files = new List<string>();
```
Youssef1313, line 41: Could this use `ImmutableHashSet.CreateBuilder`, then call `ToImmutable` at the end?

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,67 @@
...
+				foreach (var file in Context.AdditionalFiles)
+				{
+					var txt = file.GetText()?.ToString();
```
Youssef1313, line 44: This can be very expensive (noticeably), especially for large XAML files (e.g, when [XamlMerge](https://platform.uno/docs/articles/external/uno.xamlmerge.task/doc/using-xamlmerge.html) is used.

Would you be able to test the package produced in PR artifacts in a reasonable-sized project and evaluate the performance?

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,67 @@
...
+					if (txt is not null &&
+						xBindRegEx.IsMatch(txt))
```
Youssef1313, line 46: Is Regex really needed? I think it should simply be a `.Contains` check.

src/Uno.Extensions.Navigation.Generators/Uno.Extensions.Navigation.Generators.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>netstandard2.0</TargetFramework>
+		<IsPackable>false</IsPackable>
+
+		<!--
+			Defines that this project output is a tool of the given package ID,
+			so nuget override will copy it in the 'tools' directory of that package instead of the 'lib' of the current package folder.
```
Youssef1313, line 9:
```suggestion
			so nuget override will copy it in the 'analyzers' directory of that package instead of the 'lib' of the current package folder.
```

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -0,0 +1,69 @@
...
+	public IEnumerable<(string fileName, string code)> Generate()
+	{
+		var viewModels = from module in _assembly.Modules
+			from type in module.GetNamespaceTypes()
+			where IsSupported(type)
+			select type;
+
+		foreach (var vm in viewModels)
+		{
+			if (vm is not null)
```
Youssef1313, line 41: This is always true. If `NotNullWhenAttribute` is available (or can be made available), we can annotate the parameter of `IsSupported` so that a nullable warning is avoided.

src/Uno.Extensions.Navigation.Generators/Uno.Extensions.Navigation.Generators.csproj
```diff
@@ -0,0 +1,29 @@
...
+		-->
+		<ToolOfPackage>Uno.Extensions.Navigation</ToolOfPackage>
+
+		<NoWarn>$(NoWarn);RS2008</NoWarn> <!-- RS2008: Enable analyzer release tracking for the analyzer project -->
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Uno.Roslyn" PrivateAssets="All" />
```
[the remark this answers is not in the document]
Youssef1313: btw, which methods are we calling from Uno.Roslyn?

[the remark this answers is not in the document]
nickrandolph: It's probably not calling any. The references were copied from one of the other generators in the solution, so perhaps we can remove that reference.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
nickrandolph: This specific issue is only for Pages as we have to use Frame.Navigate to go to the page. For UserControls we're able to create the control and set the DataContext before adding it to the visual hierarchy (although I will double check this in navigation to verify this behaviour)

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
nickrandolph: fixed

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
nickrandolph: fixed

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
nickrandolph: fixed

src/Uno.Extensions.Navigation.Generators/Uno.Extensions.Navigation.Generators.csproj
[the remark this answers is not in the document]
nickrandolph: we're calling some of the helper methods so the reference needs to stay but I'm guessing it's optimized out of the final library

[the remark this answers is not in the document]
Youssef1313: Oh it's because it's a source package. Uno.Roslyn doesn't actually have any dlls, just plain `.cs` files in the package that gets into consuming projects.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,87 @@
...
+							{
+								hasXBind = true;
+							}
+						}
+						if(className is not null && hasXBind)
+						{
+							files.Add(className);
+						}
+					}
+				}
```
dansiegel, line 75: I'm not sure how this would be more performant that just reading the file. This seems like it would actually take more CPU cycles. The className and hasXBind can probably be moved outside of the ForEach to reduce allocations.

dansiegel (changes requested):

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,86 @@
...
+		=> IsGenerationEnabled(symbol) ?? true;
+
+	public bool? IsGenerationEnabled(ISymbol symbol)
+		=> symbol.FindAttributeValue<bool>(UpdateAttribute, nameof(ForceUpdateAttribute.IsEnabled), 0) is { isDefined: true } attribute
+			? attribute.value ?? true
+			: null;
+
+	private IImmutableSet<string>? _XBindFiles;
+	private static Regex ClassRegEx = new Regex("x:Class=\"([\\w.]+)\"");
+	private static string xBind = "{x:Bind ";
```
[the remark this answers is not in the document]
Youssef1313: const, actually :)

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
Youssef1313: It might be worth separating this into own static helper in a separate file to make the code better to deal with.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
```diff
@@ -0,0 +1,87 @@
...
+					using (var reader = new StreamReader(file.Path))
+					{
+						string? className = null;
+						var hasXBind = false;
+						while (
+							!reader.EndOfStream &&
+							(className is null ||
+							!hasXBind))
+						{
+							var txt = reader.ReadLine();
```
Youssef1313, line 54: Thinking more, this is not the most efficient code (even for large files case). It is best if we can have a fixed-size buffer that is reused instead of allocating too many small strings. It will be quite more work to do.

Say, if the buffer holds 1024 char, we will search for `{x:Bind` in it, but if the buffer ends with `{x:B` we will need to track that and start searching `ind ` in the next iteration.

Benchmarking would be good to get a feeling of what happens.

NOTE: The benchmark should be done against some .NET Framework version. I don't recall the exact version VS uses, but it won't matter much as long as it's .NET Framework.

Alternatively, we could profile the generator in a real project, instead of benchmarking.

nickrandolph: @dansiegel / @Youssef1313 this PR is accepting contributions....

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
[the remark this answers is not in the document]
dansiegel: for the string yes but for the Regex it needs to be readonly

Youssef1313 (commented):
@nickrandolph I have been thinking more on this.

Does this generator affect IntelliSense in anyway? I think the answer is no, right?

In that case, we can skip running it for design-time builds, and with that we won't care that much about performance as it will become negligible. So you can then switch back to the simple "read all the file at once" approach.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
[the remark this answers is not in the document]
nickrandolph: done

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenerationContext.cs
nickrandolph: Refactored to extract out static helper method

dansiegel (approved):

PR #1514, opened 2023-05-30
Fixing build props for IsIOS

### LINKED ISSUE #1513: Build Props set IsIOS for MacCatalyst
## Current behavior

https://github.com/unoplatform/uno.extensions/blob/1dac16670bf6838a3355516179aabb966d771fca/src/Directory.Build.props#L55

## Expected behavior

It should only be set for iOS. All checks should be netX agnostic

dansiegel:

- closes #1513 

## PR Type

- Bugfix

The check for IsIOS includes both iOS & MacCatalyst

## What is the new behavior?

The check for IsIOS only includes iOS

Android, iOS, MacCatalyst, MacOS now use netX agnostic check to prevent issues as we update from net6 in the future.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -86,7 +86,8 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-		<None Include="build\Package.targets" Pack="true" PackagePath="buildTransitive\$(PackageId).targets" Condition="exists('build')"/>
+		<None Include="build\Package.props" Pack="true" PackagePath="buildTransitive\$(PackageId).props" Condition="exists('build\Package.props')"/>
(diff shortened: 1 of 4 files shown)


src/Directory.Build.props
```diff
@@ -95,6 +96,7 @@
 	</ItemGroup>
 
 	<PropertyGroup Condition="$(_IsCatalyst)">
+		<DefineConstants>$(DefineConstants);XAMARIN</DefineConstants>
```
nickrandolph, line 99: We should correct the props so that XAMARIN is only defined for actual Xamarin targets (ie not .net6-X targets)?

nickrandolph (approved):

PR #1784, opened 2023-08-16
feat: adding maui-core fix resource paths

dansiegel:

- closes unoplatform/uno.extensions-private#103

## PR Type

- Bugfix

Assets are broken. You must reference the Resource extension from the Core project. References to `Microsoft.Maui.Controls.Application.Current.MainPage` will fail.

## What is the new behavior?

Introduces new Maui Core package which contains the extension for injecting styles.
Adds props file to disable Msix on the non-head projects (this fixes resource paths)
Initializes the Microsoft.Maui.Controls.Application with a Page (this enables scenarios where Display Alert may be called)


nickrandolph:
@mergify backport legacy/2x

PR #1803, opened 2023-08-23
Expose created Maui view on the Maui Host

dansiegel:

- n/a

## PR Type

- Feature

The Maui content is silently created by the MauiHost and is not exposed

## What is the new behavior?

The created VisualElement is now exposed through the Content property of the MauiHost. A new MauiContentCreated event has been added which can additionally be hooked onto to register a callback that should execute when the view is created.

cc: @dotmorten


src/Uno.Extensions.Maui.UI/MauiContentCreatedEventArgs.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Uno.Extensions.Maui;
+
+public class MauiContentCreatedEventArgs : EventArgs
+{
+	public MauiContentCreatedEventArgs(VisualElement content)
+	{
+		Content = content;
+	}
+
+	public VisualElement Content { get; }
```
dotMorten, line 10: Is this class necessary, over just an eventargs? You still have access to the maui view from the sender.

src/Uno.Extensions.Maui.UI/MauiHost.cs
```diff
@@ -140,6 +148,8 @@ element switch
 		};
 #endif
 
+	public new VisualElement? Content { get; private set; }
```
dotMorten, line 151: Instead of 'new' (🤮), would it be more obvious if it was a different property that made it clear it is the maui visual. Like `MauiVisual` or something.

nickrandolph: I agree, re-purposing Content is confusing. I'd suggest `MauiContent` as the property name

src/Uno.Extensions.Maui.UI/MauiHost.cs
```diff
@@ -80,7 +82,11 @@ private static void OnSourceChanged(DependencyObject dependencyObject, Dependenc
 	private static ILogger GetLogger() =>
 		IPlatformApplication.Current?.Services.GetRequiredService<ILogger<MauiHost>>() ?? throw new MauiEmbeddingInitializationException();
 
-	private VisualElement? EmbeddedView;
+	private object NativeView
```
nickrandolph, line 85: Having this property as a wrapper around base.Content is only necessary because we're overloading Content. If we change the name of the property to `MauiContent` we can remove this wrapper property too.

src/Uno.Extensions.Maui.UI/MauiContentCreatedEventArgs.cs
nickrandolph: @dotMorten the idea was to make it easier to access the VisualElement without having to cast the sender. Is there a reason why you'd prefer just an EventArgs?

src/Uno.Extensions.Maui.UI/MauiHost.cs
```diff
@@ -97,6 +103,8 @@ public MauiHost()
 		ActualThemeChanged += OnActualThemeChanged;
 	}
 
+	public event EventHandler<MauiContentCreatedEventArgs> MauiContentCreated = delegate { };
```
nickrandolph, line 106: I'd suggest `MauiContentChanged` rather than Created to be more consistent with other framework events

src/Uno.Extensions.Maui.UI/MauiContentCreatedEventArgs.cs
dotMorten: Just a slimmer api. No bigger reason than that

nickrandolph:
@mergify backport legacy/2x

nickrandolph:
@mergify backport release/stable/2.5

PR #1806, opened 2023-08-24
Initializing Application Window and Main Page

dansiegel:

- closes unoplatform/uno.extensions-private#106

## PR Type

- Bugfix

While we initialize the IApplication with a Handler, there is no Window set and the MainPage also remains uninitialized.

## What is the new behavior?

We now make sure the Application.MainPage is also set on a Window and that the Window is added to the IApplication.Windows. Both then receive a Handler with the MauiContext


nickrandolph (approved):

nickrandolph:
@mergify backport legacy/2x

PR #1814, opened 2023-08-27
fix: Enable region navigation for null request

weitzhandler:
We can change this to set the `NavigationRequestBinder` even the value is null (this PR). It looks like null or empty/whitespace string is treated the same way anyway, and will be picked up here:

https://github.com/unoplatform/uno.extensions/blob/a7151a1a06fe65fd2ebf4ebc3cf267d197d4a48c/src/Uno.Extensions.Navigation/NavigationRequestExtensions.cs#L13-L14

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/Navigation.cs
+++ b/src/Uno.Extensions.Navigation.UI/Navigation.cs
@@ -30,13 +30,13 @@ public static class Navigation
 			return;
 		}
 
-		if (d is FrameworkElement element)
+		if (d is FrameworkElement element && e.NewValue is not null)
 		{
 			_ = new NavigationRequestBinder(element);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


nickrandolph:
> We can change this to set the `NavigationRequestBinder` even the value is null (this PR). It looks like null or empty/whitespace string is treated the same way anyway, and will be picked up here:
> 
> https://github.com/unoplatform/uno.extensions/blob/a7151a1a06fe65fd2ebf4ebc3cf267d197d4a48c/src/Uno.Extensions.Navigation/NavigationRequestExtensions.cs#L13-L14

This can't work since null needs to be used to unset the dependency property. The alternative is to set a default value for the SetRequest method and for this to be picked up by the code generator

nickrandolph (approved):

Issue #1574, opened 2023-06-15
OIDC Authority issue


quintonv:
### Current behavior

I am trying out the latest UNO template and chose the OIDC auth in the wizzard. I double checked that appsettings has the settings for the hosted identity server.
When I start the mobile application in Android I just get a error on start. "No Authority Specified", I definitely have it in appsettings and in the dev version of it too.

### Expected behavior

App should start up

### How to reproduce it (as minimally and precisely as possible)

Create a new project using the Uno Platform Template

### Workaround

none

### Works on UWP/WinUI

None

### Affected platforms

Solution Templates

### IDE

Visual Studio 2022

nickrandolph:
Hi @quintonv loading from appsettings should work but to validate that there's not something else going wrong, can you try setting these values in code (`[see this example](https://github.com/unoplatform/uno.extensions/blob/1a9beaad5b02f382e8d656cd5b6ffc39278a8d8a/testing/TestHarness/TestHarness.Shared/Ext/Authentication/Oidc/OidcAuthenticationHostInit.cs#L14)`)

xperiandri:
@nickrandolph we reproduced the issue with @dim-37
Especially if we add 2 providers, then options load from the last in the `appsettings.json`
Is there a tutorial how to run tests here?
We can clone, create repro and fix if we know what is the process.

nickrandolph:
The tests for auth are all manual at the moment. You should be able to run the TestHarness app and check out the authentication pages.
With multiple providers you may need to name them so they load correctly.

nickrandolph:
@xperiandri do you have a simple repro for this that you can upload. We'll action asap and get a fix pushed

xperiandri:
@nickrandolph we prepared a fix

nickrandolph:
I'm not sure I understand what the actual problem is here (please provide a repro sample) nor what the proposed solution is trying to do.
Support for mulitple OIDC providers is done by using named config sections and calling the AddOidc multiple times.

xperiandri:
Because the options are null binding them from configuration does nothing.
@dim-37 modify the test app to have 2 providers in JSON to test that in this repo

dim-37:
Here is a test repository with the reproduced error 
https://github.com/dim-37/TestMultipleOIDC/tree/master
[...]

PR #1886, opened 2023-09-14
fix: Fix creation multiple OIDC providers from configuration

dim-37:
GitHub Issue (If applicable): #1574

## PR Type
- Bugfix

When two or more different oidc providers are created from the configuration, the settings for all of them are taken from the last specified provider.

## What is the new behavior?

Providers get the right settings when created from the configuration

## PR Checklist


src/Uno.Extensions.Authentication.Oidc/HostBuilderExtensions.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Uno.Extensions;
+﻿using Microsoft.Extensions.Configuration;
+
+namespace Uno.Extensions;
```
xperiandri, line 3: Is it by `.editorconfig` settings?

xperiandri (approved):

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
```diff
@@ -61,8 +61,11 @@ private string Generate(INamedTypeSymbol model)
 			bases: updateInterface,
 			code: $@"
 				ValueTask IForceBindingsUpdate.ForceBindingsUpdateAsync()
-				{{	
-					this.Bindings.Update();
+				{{
+					if(this.Bindings is not null)
+					{{
+						this.Bindings.Update();
+					}}
```
xperiandri, line 68: Maybe?
```suggestion
					this.Bindings?.Update();
```

src/Uno.Extensions.Authentication.Oidc/HostBuilderExtensions.cs
```diff
@@ -17,14 +19,14 @@ public static class HostBuilderExtensions
...
 
-		hostBuilder
-			.UseConfiguration(configure: configBuilder =>
-					configBuilder
-						.Section<OidcClientOptions>(name)
-				);
-
-
 		var authBuilder = builder.AsBuilder<OidcAuthenticationBuilder>();
+		if (authBuilder.Settings.Options == null)
```
nickrandolph, line 23: This block shouldn't be required, since OidcAuthenticationSettings is already registered here https://github.com/unoplatform/uno.extensions/blob/95000fa85bbf87a51c1d05ad329d96da3a0f3ac0/src/Uno.Extensions.Authentication.Oidc/HostBuilderExtensions.cs#L33

nickrandolph (changes requested):
Please provide a simple reproducible sample showing what the issue is - it would seem this can be solved using named configuration sections

xperiandri:
The fact is named config sections do not work because of this bug

dim-37:
Here is a test repository with the reproduced error
https://github.com/dim-37/TestMultipleOIDC/tree/master
In it I create 2 OIDC providers one with completely empty parameters the other with working values from TestHarness app then I try to log in with the provider with empty values and instead of getting an error due to empty values it uses the values of the last registered provider in this case the provider with values from the TestHarness app.

src/Uno.Extensions.Navigation.Generators/ForceBindingsUpdateGenTool_1.cs
nickrandolph: Is there a scenario where this is being called with this.Bindings is null?

nickrandolph:
This will be replaced by this https://github.com/unoplatform/uno.extensions/pull/1889

PR #1933, opened 2023-10-04
Use unified container between Uno & Maui

dansiegel:

- Replaces ##1844

## PR Type

- Feature

Uno's Host Builder and the MauiAppBuilder provide 2 separate instances of the IServiceProvider each with their own registered services. This means that services registered in one cannot be properly resolved by the other.

## What is the new behavior?

Both builders have an IServiceProviderFactory. When using the IHostBuilder the initialization of the MauiAppBuilder is deferred so that all Services can be registered with both the HostBuilder's IServiceCollection and the MauiAppBuilder.Services. The two are combined and an IServiceProvider is created. This IServiceProvider is then shared with .NET MAUI so that both Uno and .NET MAUI can make use of the same container with all of the same services available to each.

**NOTE** In testing the only "Duplicated" Service out of the box is IConfiguration. To prevent this from being an issue when we combine the two IServiceCollection's we use TryAdd. This prevents the addition of a ServiceDescriptor from the .NET MAUI side with a ServiceType that has already been provided on the Uno side.

### DIFF
--- a/src/Uno.Extensions.Maui.UI/UnoHostBuilderServiceProviderFactory.cs
+++ b/src/Uno.Extensions.Maui.UI/UnoHostBuilderServiceProviderFactory.cs
@@ -1,0 +1,30 @@
+#if MAUI_EMBEDDING
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Uno.Extensions.Maui;
+internal class UnoHostBuilderServiceProviderFactory : IServiceProviderFactory<IServiceProvider>
+{
+	private readonly MauiAppBuilder _mauiAppBuilder;
+	private readonly Action _buildAppBuilderCallback;
+
+	public UnoHostBuilderServiceProviderFactory(MauiAppBuilder builder, Action buildAppBuilderCallback)
+	{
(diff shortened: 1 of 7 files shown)


src/Uno.Extensions.Maui.UI/UnoHostBuilderServiceProviderFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+	{
+		_mauiAppBuilder = builder;
+		_buildAppBuilderCallback = buildAppBuilderCallback;
+	}
+
+	// We delay calling build on the MauiHostBuilder until the Uno Host is built
+	public IServiceProvider CreateBuilder(IServiceCollection services)
+	{
+		// NOTE: TryAdd prevents duplicates from overriding what was registered with Uno's Host Builder such as IConfiguration
+		_mauiAppBuilder.Services.ForEach(x => services.TryAdd(x));
```
nickrandolph, line 20: Won't this result in two instances of singletons, one in each container?

nickrandolph (approved):

Issue #1605, opened 2023-06-29
[MVUX] Better support for nullability annotations


dr1rrb:
## What would you like to be added:
`Feed.Async<T>` where `T : notnull` should accept an `AsyncFunc<T?>` and automatically interpret a `null` as `None`.

Same for other operators like `.Select`, and `.SelectAsync`.

## Why is this needed:
The language now have the nullable annotations which are commonly used by devs. Doing such would improve user experience and help to use the `None` state which is often not used.

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] WebAssembly renders for Xamarin.Forms
- [x] Windows
- [ ] Build tasks

## Anything else we need to know?
* The `Option<T>` should be hidden from "first level public API" (could be usable only with "advanced" operators)
* We already have such support for `State.Update`
* This could be a breaking change (compile time)

nickrandolph:
@dr1rrb is this something we can get done in 5.1? If so, can you update the release number

dr1rrb:
When we look at "operators", we have 2 cases to consider:
1. The input type parameter (`Feed.Select`; `Feed.SelectAsync`; `Feed.Where`; `State.Update`; `State.ForEachAsync`)
2. The return type (`Feed.Select`; `Feed.SelectAsync`; `State.Update`)

Case 1 could be named "None as null", while case 2 is "null as None".

# Proposal
To match the "new behavior" suggested for "factories", we should also follow the "null as None". This means that we should constraint the "return type" on all operators to `notnull` and convert the `null` to `None` (still following the feature flag).

In order to allow users to return `Some(null)` we would also have to add `.SelectData(Func<Option<TSource>, Option<TResult>>)` (using the `Data` suffix like we already have to `State.Data`), `SelectDataAsync` and `WhereData`.

However we should NOT implement the "None as null", this is to avoid to implicitly convert `None(struct)` to `Some(default(struct))` (e.g. having the "None as `null`" would allow `Feed<int>.Select(i => i + 1)` which would implicitly convert `None<int>()` to `Some(0)`)

# Special case for the State
Currently `State.Update` and `State.ForEachAsync` are already constraints to `notnull` on "input parameter". 

## State.Update
Input and output types are the same, implementing the "null as None" somehow forces us to also implement the "None as null". 

This is not as problematic as for `Feed.Select` as the `Update` is an imperative operation where we anyway expect to change the current value. Getting a default instead of `None` is actually the expected behavior.

[...]

PR #2080, opened 2023-11-29
feat: [MVUX] Improve nullable support

dr1rrb:
closes https://github.com/unoplatform/uno.extensions/issues/1605

## Feature
Improve nullable support in feed factories and relevant operators

Nullable annotation are not used to interpreted by the feeds. There is no recommendation to use non nullable types in feeds and users are often involuntary using using `Some(null)` while they should use (/ thinking they are using) `None`.

## What is the new behavior?
* Most common factory methods (`<Feed|State>.Async[Enumerable]`) are now constrained with `T : notnull`
* `Select` and `SelectAsync` operators are also constrained with `TResult : notnull`
* When user returns a `T`  with the `T : struct` constraint, we allowed him to return a `T?` which is interpreted as `None`
* In any time, when user provide a `T` that is ref type and `null`, it's interpreted as `None` (e.g. `Feed<MyObject>.Async(async _ => null!)` will now goes in `None`).
* `SelectData` and `SelectDataAsync` operators has been added to allow user to still use `Some(null)` if needed.

## PR Checklist
- [ ] ~~Contains **NO** breaking changes~~

## Other information
# COMPILATION BREAKING CHANGES
This PR introduces few compilation breaking changes. Here is a shot summary of the known breaking 
* `Feed.Async(async _ => default(MyObject))` is no longer allowed, to fix you need to specify the type of the value of the feed, i.e. use `Feed**<MyObject?>**.Async(async _ => default(MyObject))`. Same for `Feed.AsyncEnumerable`, `State.Async` and `State.AsyncEnumerable`
* `myFeed.Select(_ => default(MyObject))` is no longer allowed, to fix you need to use the `SelectData` operator. Same for `SelectAsync` where you need use `SelectDataAsync`.

# BEHAVIOR BREAKING CHANGES
This PR also introduces few behavior changes that could be breaking in few use case. Those are describe in the issue https://github.com/unoplatform/uno.extensions/issues/1605, and which can be summarized "any time you return a `null` value, it's now automatically converted as `None`". This was already the case for some factories and operators and is expected to cause issue only in really rare use-cases. If you are encountering such an issue, use factories and operators overload that accepts `Option<T>` instead of `T`.


nickrandolph:
@dr1rrb how are we going to deal with the breaking changes. I assume we need to push to v4 of extensions, to reflect breaking changes?

dr1rrb:
> @dr1rrb how are we going to deal with the breaking changes. I assume we need to push to v4 of extensions, to reflect breaking changes?

I would agree, but usually (unfortunately) the "major" version number is more often a "marketing number" than a "semantic number"... especially since we have released v3 quite recently. We will have to discuss about that ...

PR #2215, opened 2024-03-25
Updated infrastructure packages

xperiandri:
## PR Type

- Code style update (formatting)
- Refactoring (no functional changes, no API changes)

## PR Checklist


src/Directory.Packages.props
```diff
@@ -4,11 +4,12 @@
 		<PackageVersion Include="DotNet.ReproducibleBuilds" Version="1.1.1" />
 		<PackageVersion Include="FluentAssertions" Version="6.7.0" />
 		<PackageVersion Include="IdentityModel.OidcClient" Version="5.0.0" />
+		<PackageVersion Include="FluentValidation " Version="11.9.0" />
 		<PackageVersion Include="MSTest.TestAdapter" Version="2.2.9" />
 		<PackageVersion Include="MSTest.TestFramework" Version="2.2.9" />
 		<PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
-		<PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" />
-		<PackageVersion Include="Microsoft.Extensions.Configuration.Abstractions" Version="8.0.0"  />
+		<PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.9.2" />
```
Youssef1313, line 11: Why is this update needed?
It will require Visual Studio 17.9, so 17.8 and earlier will no longer be supported.

src/Directory.Packages.props
```diff
@@ -20,10 +21,10 @@
 		<PackageVersion Include="Microsoft.Maui.Controls.Compatibility" Version="8.0.3" />
 		<PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.6.2" />
 		<PackageVersion Include="Microsoft.Net.Compilers.Toolset" Version="4.2.0" />
-		<PackageVersion Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
+		<PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
 		<PackageVersion Include="Microsoft.UI.Xaml" Version="2.7.1" />
-		<PackageVersion Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.756" />
-	    <PackageVersion Include="Microsoft.WindowsAppSDK" Version="1.4.231219000" />
+		<PackageVersion Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.3233" />
+		<PackageVersion Include="Microsoft.WindowsAppSDK" Version="1.5.240311000" />
```
Youssef1313, line 27: Not sure if we want to bump this for now. We had a discussion with @nickrandolph around some issues that are "partially" on WinUI side and are fixed in 1.5, but I'm not sure if we'll end up wanting to update or not.

src/Uno.Extensions.Core.Generators/Compat/CompatibilityTypesGenerationTool.cs
```diff
@@ -402,7 +402,9 @@ private bool GetIsDisabled(string propertyName)
 		}
 		using var resourceReader = new StreamReader(resource);
 
+#pragma warning disable RS1035 // Using Environment.NewLine is fine
```
Youssef1313, line 405: Fine for now, but not yet clear if we should be using `Environment.NewLine` or not. There is a bit of discussion around that in https://github.com/dotnet/roslyn-analyzers/issues/6467

src/Directory.Packages.props
xperiandri: I thought it just brings new code analysis features

Youssef1313: @xperiandri Nope, this has an impact on the minimum supported VS version

PR #2397, opened 2024-06-26
feat: Add `doNotActivate` parameter to `NavigateAsync` method from Navigation Extensions

### LINKED ISSUE #2381: [ApplicationBuilder] must not create Window until it ensures that the host started

Trying to use OpenIddict Client it inserts a special `IHostedService` that extracts protocol activation, redirects activation and stops the application.
This means until host initialization is completed Windows must not be created as initialization can probably fail or be terminated.

Now a Window always opens and closes after some period of time if that OpenIddict Client hosted service extracts activation.

## Why is this needed:

To prevent Windows creation in unnecessary situations.

## For which Platform:

- Windows
- Desktop

xperiandri:
GitHub Issue (If applicable): closes #2381

## PR Type

- Feature

Window is always activated as soon as possible

## What is the new behavior?

Window is activated either as soon as possible or right after the host is started (all hosted services completed)

## PR Checklist

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/IRootViewInitializer.cs
+++ b/src/Uno.Extensions.Navigation.UI/IRootViewInitializer.cs
@@ -22,5 +22,6 @@ internal interface IRootViewInitializer
 	/// <param name="window"></param>
 	/// <param name="element"></param>
 	/// <param name="loadingTask"></param>
-	void InitializeViewHost(Window window, FrameworkElement element, Task loadingTask);
+	/// <param name="doNotActivate"></param>
+	void InitializeViewHost(Window window, FrameworkElement element, Task loadingTask, bool doNotActivate = false);
 }
--- a/src/Uno.Extensions.Navigation.Toolkit/ApplicationBuilderExtensions.cs
+++ b/src/Uno.Extensions.Navigation.Toolkit/ApplicationBuilderExtensions.cs
@@ -6,9 +6,9 @@ public static class ApplicationBuilderExtensions
 	{
 		public ContentControl CreateDefaultView() => new LoadingView();
 
-		public void InitializeViewHost(Window window, FrameworkElement element, Task loadingTask)
+		public void InitializeViewHost(Window window, FrameworkElement element, Task loadingTask, bool doNotActivate)
 		{
-			window.ApplyLoadingTask(element, loadingTask);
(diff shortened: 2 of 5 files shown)


nickrandolph:
@xperiandri can you provide more context on the scenario you're trying to get to work?

xperiandri:
OpenIddict Client starts a hosted service to determine protocol activation, redirect it and close the app.
With you current logic the second instance of the app manages to activate the window before the app exited. It causes a window to blip and close.
My modification allows to prevent that if necessary

nickrandolph:
ok let me double check this next week - I thought we had a mechanism that would ensure some services are started before activation (eg localization from recollection).
If we do want to provide this, we'll possibly look at providing it as a registration on the DI, rather than as an argument to the app builder. 
Thanks again for this suggestion!

xperiandri:
I need this behavior, please

xperiandri:
And I would prefer ability to just plug OpenIddict Client and it worked then declaring some additional services, implement some additional interfaces, etc.

src/Uno.Extensions.Navigation.UI/IRootViewInitializer.cs
```diff
@@ -22,5 +22,6 @@ internal interface IRootViewInitializer
 	/// <param name="window"></param>
 	/// <param name="element"></param>
 	/// <param name="loadingTask"></param>
-	void InitializeViewHost(Window window, FrameworkElement element, Task loadingTask);
+	/// <param name="doNotActivate"></param>
+	void InitializeViewHost(Window window, FrameworkElement element, Task loadingTask, bool doNotActivate = false);
```
carldebilly, line 26: That's odd using a _double negative_ like that. That's confusing.

carldebilly (approved):

nickrandolph:
@xperiandri the interface I was thinking of is the IStartupService, which will effectively block activation until it returns. You can register this service in order to block activation.
I'd like to explore what we can do to make it eaiser to just "plug OpenIDdict client" and leverage the authentionservice.

xperiandri:
Sounds like a good idea in general.
The only question is how to make https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Client.SystemIntegration/OpenIddictClientSystemIntegrationActivationHandler.cs become `IStartupService`

Inherit from it, remove registration as IHostedService from DI and register as `IStartupService`?

PR #2425, opened 2024-07-12
feat: Implement `ViewMapNotFoundException`s

### LINKED ISSUE #2424: Missing ViewMap produces meaningless exception complicated to figure out promblem instantly

`ViewMapNotFoundException`

## Why is this needed:

Hard to figure out what is wrong on runtime

xperiandri:
GitHub Issue : closes #2424 

## PR Type

- Feature

Meaningless exception

## What is the new behavior?

`ViewMapNotFoundException` derived exceptions

## PR Checklist

- [X] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md). (for bug fixes / features) 

### DIFF
--- a/src/Uno.Extensions.Navigation/ViewMapNotFoundException.cs
+++ b/src/Uno.Extensions.Navigation/ViewMapNotFoundException.cs
@@ -1,0 +1,141 @@
+﻿namespace Uno.Extensions.Navigation;
+
+/// <summary>
+/// Represents an exception that is thrown when a view map is not found in the view registry.
+/// </summary>
+public class ViewMapNotFoundException : Exception
+{
+	/// <summary>Initializes a new instance of the <see cref="ViewMapNotFoundException"></see> class.</summary>
+	private ViewMapNotFoundException() { }
+
+	/// <summary>Initializes a new instance of the <see cref="ViewMapNotFoundException"></see> class with a specified error message.</summary>
+	/// <param name="message">The message that describes the error.</param>
+	public ViewMapNotFoundException(string message) : base(message) { }
+
+	/// <summary>Initializes a new instance of the <see cref="ViewMapNotFoundException"></see> class with a specified error message and a reference to the inner exception that is the cause of this exception.</summary>
+	/// <param name="message">The error message that explains the reason for the exception.</param>
+	/// <param name="innerException">The exception that is the cause of the current exception, or a null reference (Nothing in Visual Basic) if no inner exception is specified.</param>
+	public ViewMapNotFoundException(string message, Exception innerException) : base(message, innerException) { }
+}
+
+/// <summary>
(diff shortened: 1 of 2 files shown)


src/Uno.Extensions.Navigation/ViewMapNotFoundException.cs
```diff
@@ -0,0 +1,141 @@
+﻿namespace Uno.Extensions.Navigation;
+
+/// <summary>
+/// Represents an exception that is thrown when a view map is not found in the view registry.
+/// </summary>
+public class ViewMapNotFoundException : Exception
```
eriklimakc, line 6: Hello @xperiandri, thanks for the PR! Maybe we could simplify a bit and have something like:

```csharp
public enum ViewMapSearchType
    View,
    ViewModel,
    Data,
    ResultData

public class ViewMapNotFoundException : Exception

    public ViewMapNotFoundException(ViewMapSearchType searchType, Type? type = null)
        : base(CreateNotFoundMessage(searchType, type))

    public ViewMapNotFoundException(ViewMapSearchType searchType, Type? type, Exception innerException)
        :base(CreateNotFoundMessage(searchType, type), innerException)

    private static string CreateNotFoundMessage(ViewMapSearchType searchType, Type? type)
```

Or if it's preferable to have multiple exception types (so they can be caught independently), we could define a base exception class like `ViewMapNotFoundExceptionBase`. In this base class, we can define the `CreateTypeNotFoundMessage` method such as:

```csharp
CreateTypeNotFoundMessage(string typeName, Type? type) 
    => $"The {typeName} type {type} was not found in the view registry.";
```

This method can then be used by the other exception classes that inherit from `ViewMapNotFoundExceptionBase`. Remember to use the `enum ViewMapSearchType`, please.

eriklimakc (commented):
Could you please update your fork with the latest changes from the main repo? It includes an update to the `System.Text.Json` package to 8.0.4. The previous version 8.0.1 now has vulnerabilities.

src/Uno.Extensions.Navigation/ViewMapNotFoundException.cs
xperiandri: I defined a base class.
Type comes from different errors, so different exception types needed

xperiandri:
Repo contributor can rebase my branch onto the latest dev with 2 clicks

francoistanguay:
Why do we need a specific exception type? I dont see Try/Catch happening on that specific exception type being useful. Couldnt we just have a better error message?

Ideally those discussions would happen at the "specification" level on a discussion first please so there's less back and forth.

xperiandri:
I will go that way next time

xperiandri:
So as I mentioned before, if you forget to register some mapping you will see ArgumentException sequence contains no elements. It means nothing to understand what was the issue

PR #2493, opened 2024-08-02
chore: Fixing canary with updated nuget updater

nickrandolph:


testing/TestHarness/TestHarness/TestHarness.csproj
```diff
@@ -76,17 +76,19 @@
 		<ProjectReference Include="..\TestHarness.Core\TestHarness.Core.csproj" />
 
 		<PackageReference Include="Uno.Core.Extensions.Logging.Singleton"/>
+		<PackageReference Include="Uno.WinUI" />
+		<PackageReference Include="Uno.UI.Adapter.Microsoft.Extensions.Logging"/>
```
Youssef1313, line 80: Shouldn't these be added implicitly by Uno.Sdk?

nickrandolph: They need to be explicit in order for the canary updater to work as it was complaining about version mismatch

Youssef1313 (approved):

PR #2517, opened 2024-08-21
fix: App blanks out when navigating to invalid route

### LINKED ISSUE #2382: App Layout Blanks Out When Navigating to Invalid Route
## Current behavior
When navigating to a route that matches a common type name with suffixes (e.g., List, resulting in ListView), the app goes blank and logs a warning: Uno.Extensions.Navigation.Navigators.ContentControlNavigator: Warning: CreateControlFromType - Type not registered ListView, so calling Activator.

## Expected behavior
Navigating to a route that is not registered should handle the error and avoid causing the app to go blank/crash.

1. Create a new Uno App.
2. Set up a navigation button with an invalid path using uen:Navigation.Request (e.g., List).
3. Observe that navigating to this invalid path causes the app layout to go blank.
Minimal Repro Project: [NavigationBug.zip](https://github.com/user-attachments/files/15929129/NavigationBug.zip)

## Environment

Nuget Package (s):

Package Version(s):

Affected platform(s):

- [x] iOS
- [ ] macOS (AppKit)
- [ ] Mac Catalyst
- [x] Android
- [ ] WebAssembly
- [x] Windows
- [ ] Skia (WPF)
- [ ] Skia (GTK on Linux/macOS/Windows)
- [ ] Skia (Linux Framebuffer)
- [ ] Build tasks

Visual Studio:

- [ ] 2019 (version: )
- [x] 2022 (version: )
- [ ] Visual Studio Code (version: )
- [ ] Rider Windows (version: )
- [ ] Rider macOS (version: )

Relevant plugins:

- [ ] Resharper (version: )

 Original Issue : https://github.com/unoplatform/uno/issues/17238
 
The issue seems to stem from the type resolution logic in RouteResolverDefault: https://github.com/unoplatform/uno.extensions/blob/main/src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs#L149-L173
https://github.com/unoplatform/uno.extensions/blob/848ac33909af8b4f5c14fd42ed3f7424f23350ed/src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs#L9

vatsashah45:
GitHub Issue (If applicable): closes #2382 

## PR Type

- Bugfix

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs
+++ b/src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs
@@ -124,11 +124,8 @@ public class RouteResolverDefault : RouteResolver
 
 			if (view is not null)
 			{
-				if (view != null)
-				{
-					if (Logger.IsEnabled(LogLevel.Information))
-						Logger.LogInformationMessage($"Potential conflict detected: The route '{path}' resolved to a common control or class name '{view.Name}'. This could lead to unexpected behavior.");
-				}
+				if (Logger.IsEnabled(LogLevel.Information))
+					Logger.LogInformationMessage($"Potential conflict detected: The route '{path}' resolved to a common control or class name '{view.Name}'. This could lead to unexpected behavior.");
 			}
 			
 		}
--- a/src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs
+++ b/src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs
@@ -82,6 +82,8 @@ public class RouteResolverDefault : RouteResolver
 		if (path is null ||
 			string.IsNullOrWhiteSpace(path))
 		{
+			if (Logger.IsEnabled(LogLevel.Warning))
+				Logger.LogWarningMessage($"Unable to resolve path from types. Path: '{path}', View: '{view?.Name}', ViewModel: '{viewModel?.Name}'");
 			return default;
 		}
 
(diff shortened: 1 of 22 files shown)


eriklimakc (approved):

src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs
```diff
@@ -130,6 +132,12 @@ view is null &&
 			viewModel = TypeFromPath(trimmedPath, false, ViewModelSuffixes);
 		}
 
+		if (view != null && IsCommonControlName(view.Name))
+		{
+			if (Logger.IsEnabled(LogLevel.Warning))
+				Logger.LogWarningMessage($"Potential conflict detected: The route '{path}' resolved to a common control or class name '{view.Name}'. This could lead to unexpected behavior.");
+		}
```
kazo0, line 139: I'm not sure I understand this check here. It is still a valid scenario to be able to navigate to a known FrameworkElement. I think we should change this to an Information level log at least. 

And I would get rid of the IsCommonControlName logic. We should set some sort of bool after line 126 to check if view is not null. That would guarantee that we have found a known FrameworkElement. We can use that bool to then log an informational message explaining that the route has been resolved to a known type. Maintaining a list of "Common" control names could be fragile as we may not cover every single scenario.

src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs
```diff
@@ -143,13 +151,21 @@ view is null &&
 		}
 
 		if (Logger.IsEnabled(LogLevel.Warning)) Logger.LogWarningMessage($"Unable to create default mapping");
-		return default;
+		return null;
```
kazo0, line 154:
```suggestion
		return default;
```

src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs
```diff
@@ -96,32 +94,43 @@ view is null &&
...
 
 		if (view is null)
 		{
 			var trimmedPath = TrimSuffices(path, ViewModelSuffixes);
 			view = TypeFromPath(trimmedPath, true, ViewSuffixes, type => type.IsSubclassOf(typeof(FrameworkElement)));
+
+			if (view is not null)
+			{
+				if (view != null)
+				{
```
kazo0, line 128: You don't need this check, you are already inside of the `view is not null` if block

kazo0 (approved):

PR #2586, opened 2024-10-12
Update WorkingWithRecords.md

JohannesMP:
Message field `status` is declared lowercase, but was used uppercase.

### DIFF
--- a/doc/Learn/Mvux/WorkingWithRecords.md
+++ b/doc/Learn/Mvux/WorkingWithRecords.md
@@ -106,7 +106,7 @@ In our Model:
 message = message with
 {
     Content = response.Message,
-    Status = response.IsError ? Status.Error : Status.Value
+    status = response.IsError ? Status.Error : Status.Value
 };
 
 //Then you can update your message list displayed in the UI, thread-safe


ajpinedam:
Thank you @JohannesMP  for your contribution!

ajpinedam (approved):

ajpinedam:
@JohannesMP, could you please correct the commit message so it matches the convention?  https://www.conventionalcommits.org/en/v1.0.0/

Since this is a documentation change, all you need is to add `docs:` at the beginning, something like:

```bash
docs: Update WorkingWithRecords.md
```

doc/Learn/Mvux/WorkingWithRecords.md
```diff
@@ -106,7 +106,7 @@ In our Model:
 message = message with
 {
     Content = response.Message,
-    Status = response.IsError ? Status.Error : Status.Value
+    status = response.IsError ? Status.Error : Status.Value
```
kazo0, line 109: I think it would be best instead to alter the property declarations here: https://github.com/JohannesMP/uno.extensions/blob/5e2e4a67c36c9cfef6fcbf21a2d393166334ed35/doc/Learn/Mvux/WorkingWithRecords.md?plain=1#L98

And change it to properly capitalize each of those public properties on the `record`. So it should be 

```csharp
public partial record Message(string Content, Status Status, Source Source);
```

PR #2623, opened 2024-11-19
chore: use sdk props for dependencies

kazo0:
Fixing canaries as we are getting package downgrade errors when bumping to latest uno.sdk


nickrandolph:
@kazo0 might be worth switching to using the uno.sdk.private as part of this update

eriklimakc (approved):

PR #2624, opened 2024-11-20
fix: Reduce winappsdk version dependency

nickrandolph:
This is required in order to prevent version conflicts with uno.sdk which is rolling back winappsdk version to the latest version currently supported by the Store
See: https://github.com/microsoft/WindowsAppSDK/issues/4480 is resolved


jeromelaban (approved):

nickrandolph:
@mergifyio backport release/stable/5.1

PR #2659, opened 2025-01-10
feat: Add back support for MSAL auth through WAM

### LINKED ISSUE #2654: [Authentication][MSAL] Add back support for auth through WAM
> Afaik opening the account picker in the default browser is the recommended approach as per the latest security standards (instead of the embedded browser) - the docs say so quite overwhelmingly as well - https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/using-web-browsers#embedded-web-view-vs-system-browser

Correct for a browser based approach - however the desired and recommended behavior on Windows is to use the native WAM, which is built into the OS:
https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/desktop-mobile/wam

Using the native Windows WAM was the behavior up to Uno.Sdk 5.2.139, which was broken in 5.2.175; that is what I referred to in this issue:
> Expected behavior
I can update to the latest stable Uno Sdk and continue to use the native account picker on Windows with the Uno Auth Extensions.

_Originally posted by @VincentH-Net in https://github.com/unoplatform/uno.extensions/issues/2443#issuecomment-2540952969_

Allow the MSAL Auth Extensions to be configured to use the WAM (https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/desktop-mobile/wam). This will require a change to the IApplicationBuilder flow so that the instance of the main application Window can be retrieved and passed to [.WithParentActivityOrWindow](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/desktop-mobile/wam#parent-window-handles)

eriklimakc:
GitHub Issue (If applicable): closes #2654

## PR Type

- Feature

## PR Checklist

## Other information

### DIFF
--- a/doc/Learn/Authentication/HowTo-MsalAuthentication.md
+++ b/doc/Learn/Authentication/HowTo-MsalAuthentication.md
@@ -48,7 +48,7 @@ uid: Uno.Extensions.Authentication.HowToMsalAuthentication
     }
     ```
 
-- Add the `MsalAuthenticationProvider` using the `AddMsal()` extension method which configures the `IAuthenticationBuilder` to use it.
+- Use the `Configure` method overload that provides access to a `Window` instance. Add the `MsalAuthenticationProvider` using the `AddMsal()` extension method which configures the `IAuthenticationBuilder` to use it.
 
     ```csharp
     private IHost Host { get; set; }
@@ -56,17 +56,20 @@ uid: Uno.Extensions.Authentication.HowToMsalAuthentication
     protected override void OnLaunched(LaunchActivatedEventArgs args)
     {
         var builder = this.CreateBuilder(args)
-            .Configure(host =>
+            .Configure((host, window) =>
             {
                 host
                 .UseAuthentication(builder =>
                 {
-                    builder.AddMsal();
+                    builder.AddMsal(window);
                 });
             });
         ...
     }
     ```
+
+> [!IMPORTANT]
+> The `AddMsal()` method requires a `Window` instance, which the `MsalAuthenticationProvider` uses to set up the authentication dialog. You can access the `Window` instance through the `Configure()` method overload that provides it.
 
 - The `IAuthenticationBuilder` is responsible for managing the lifecycle of the associated provider that was built.
 
@@ -95,12 +98,12 @@ uid: Uno.Extensions.Authentication.HowToMsalAuthentication
     protected override void OnLaunched(LaunchActivatedEventArgs args)
     {
         var builder = this.CreateBuilder(args)
-            .Configure(host =>
+            .Configure((host, window) =>
             {
                 host
                 .UseAuthentication(builder =>
                 {
-                    builder.AddMsal(msal =>
+                    builder.AddMsal(window, msal =>
                         msal
                         .Builder(msalBuilder => 
                             msalBuilder.WithClientId("161a9fb5-3b16-487a-81a2-ac45dcc0ad3b"))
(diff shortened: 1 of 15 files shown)


kazo0:
@Mergifyio backport release/stable/5.2

src/Uno.Extensions.Authentication.MSAL/MsalAuthenticationProvider.cs
```diff
@@ -1,7 +1,8 @@
 ﻿#if WINDOWS
-using Microsoft.Identity.Client.Desktop;
+using Microsoft.Identity.Client.Broker;
 #endif
 using Uno.Extensions.Logging;
+using WinRT.Interop;
```
kazo0, line 5: This should probably be within the `#if WINDOWS`

src/Uno.Extensions.Authentication.MSAL/MsalAuthenticationProvider.cs
```diff
@@ -50,7 +51,12 @@ public void Build()
 		}
 
 #if WINDOWS
-		builder.WithWindowsEmbeddedBrowserSupport();
+		builder.WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows));
+		builder.WithParentActivityOrWindow(() =>
+		{
+			IntPtr hwnd = WindowNative.GetWindowHandle(window);
```
kazo0, line 57: Actually you coudl remove the `using WinRT.Interop` and just fully qualify this to `WinRT.Interop.InitializeWithWindow` like we do here: https://github.com/unoplatform/uno.extensions/blob/78184f87453578f3fa6f9a9b1275ada1ef7b7fe6/src/Uno.Extensions.Navigation.UI/Navigators/MessageDialogNavigator.cs#L101

src/Uno.Extensions.Authentication.MSAL/HostBuilderExtensions.cs
```diff
@@ -21,11 +21,48 @@ public static class HostBuilderExtensions
 	/// <returns>
 	/// The <see cref="IAuthenticationBuilder"/> that was passed in.
 	/// </returns>
+	[Obsolete("This method is obsolete. Please use the AddMsal overload that accepts a 'Window' parameter to specify the authentication window. The overload without 'Window' will be removed in a future release.", false)]
```
kazo0, line 24: We'll need to edit the current docs: https://github.com/unoplatform/uno.extensions/blob/78184f87453578f3fa6f9a9b1275ada1ef7b7fe6/doc/Learn/Authentication/HowTo-MsalAuthentication.md?plain=1#L51

Not sure if we have anywhere that we are documenting the upgrade experience from one version to another for Extensions. Similar to things like Toolkit: https://platform.uno/docs/articles/external/uno.toolkit.ui/doc/material-migration.html

Though it's mostly for documenting upgrading to new major versions. Either way, let's make sure we do a pass where this should be added/called out in the documentation

src/Uno.Extensions.Authentication.MSAL/MsalAuthenticationProvider.cs
eriklimakc: replaced by https://github.com/unoplatform/uno.extensions/pull/2659#discussion_r1910495906

doc/Learn/Authentication/HowTo-MsalAuthentication.md
```diff
@@ -48,26 +48,29 @@ uid: Uno.Extensions.Authentication.HowToMsalAuthentication
...
-                    builder.AddMsal();
+                    builder.AddMsal(window);
                 });
             });
         ...
     }
     ```
 
+> [!IMPORTANT]
+> The `AddMsal()` method requires a `Window` instance, which the `MsalAuthenticationProvider` uses to set up the authentication dialog. You can access the `Window` instance through the `Configure()` method overload that provides it.
```
kazo0, line 72: I don't think it'd be required, its only there so the WAM will display properly. Omitting it would fall back to the deprecated method but it wouldnt cause an error

kazo0 (approved):

PR #2737, opened 2025-04-23
fix: add region routes for panel vis navigator on loaded

### LINKED ISSUE #2690: [Navigation][TabBar][NavView] Avoid using reflection when navigating through tabs with a region attached

When navigating using a TabBar or NavigationView with a `Grid` to hold the content of its items, it would be great if reflection was not used, since the items are already there.

As an example we have the code shown in the docs:

[NavigationView](https://platform.uno/docs/articles/external/uno.extensions/doc/Learn/Navigation/Advanced/HowTo-UseNavigationView.html#code-example
)
[TabBar](https://platform.uno/docs/articles/external/uno.extensions/doc/Learn/Navigation/Advanced/HowTo-UseTabBar.html#code-example
)

If there are no RouteMaps for the tabs, when preparing the request to navigate Navigation goes searching for a [view ](https://github.com/unoplatform/uno.extensions/blob/main//src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs#L123)or [view model](https://github.com/unoplatform/uno.extensions/blob/main//src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs#L136) to dynamically create a RouteInfo for that navigation. But there would be no views nor VMs associated there. So in the end it ends up creating a RouteInfo with only the path [here](https://github.com/unoplatform/uno.extensions/blob/main//src/Uno.Extensions.Navigation.UI/RouteResolverDefault.cs#L141-L144).

## For which Platform:

- [X] All

kazo0:
closes #2690

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/Navigators/PanelVisiblityNavigator.cs
+++ b/src/Uno.Extensions.Navigation.UI/Navigators/PanelVisiblityNavigator.cs
@@ -14,21 +14,32 @@ public class PanelVisiblityNavigator : ControlNavigator<Panel>
 		RegionControlProvider controlProvider)
 		: base(logger, dispatcher, region, resolver, controlProvider.RegionControl as Grid)
 	{
-		if (region.View is not null)
+		if (region.View is { } view)
 		{
-			region.View.Loaded += PanelLoaded;
+			if (view.IsLoaded)
+			{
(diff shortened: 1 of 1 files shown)


kazo0:
@Mergifyio backport release/stable/5.3

PR #2786, opened 2025-05-30
fix: Use initial Region to collect full route

### LINKED ISSUE #2771: [Navigation] Address bar doesn't update for overlay pages on Tabs/NavView
Regular navigation on a Tabs/NavViewItems update the address bar accordingly. However, if a page is navigated where it is open on top of the TabBar or NavView, the address bar won't update and when navigating back the address bar only updates once.

address-bar-issue2

### Investigation details:

The reason behind the "Settings" route not being added to the addressbar is because somewhere after the navigation is started the Navigator (NavigationViewNavigator) loses its children and its parent. 
So when it comes to the end of the navigation [here](https://github.com/unoplatform/uno.extensions/blob/1d82f054d12908ced1ec23924dee0cafb0b7434e/src/Uno.Extensions.Navigation.UI/Navigator.cs#L96), it has lost them and then when we try to build the full path to add to the addressbar it returns null because it relies on the `Region.Root()` - that in our case is the own `NavigationViewNavigator` since he has no Parent.

_Originally posted by @erikvilima in https://github.com/unoplatform/uno.extensions/issues/2768#issuecomment-2880659198_

erikvilima:
GitHub Issue (If applicable): closes #2771

## PR Type

- Bugfix

Because of `ViewUnloaded` timing being different on WASM by the time we update the addressbar the Regions have lost its children/parent.

## What is the new behavior?
Use the `Region` from the beginning of the navigation to get the full route.

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/RouteNotifier.cs
+++ b/src/Uno.Extensions.Navigation.UI/RouteNotifier.cs
@@ -16,12 +16,12 @@ internal class RouteNotifier : IRouteNotifier, IRouteUpdater
 
 	private IDictionary<Guid, StringBuilder> navigationSegments = new Dictionary<Guid, StringBuilder>();
 	private IDictionary<Guid, int> runningNavigations = new Dictionary<Guid, int>();
-	private IRegion? _rootRegion;
+	private IDictionary<Guid, IRegion> rootRegions = new Dictionary<Guid, IRegion>();
 
 	public void StartNavigation(INavigator navigator, IRegion region, NavigationRequest request)
 	{
 		var id = request.Id;
-		_rootRegion = region.Root();
+		rootRegions[id] = region.Root();
 
 		if (!runningNavigations.TryGetValue(id, out var count) ||
 			count == 0)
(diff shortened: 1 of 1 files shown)


src/Uno.Extensions.Navigation.UI/RouteNotifier.cs
```diff
@@ -17,9 +17,11 @@ public RouteNotifier(ILogger<RouteNotifier> logger)
 	private IDictionary<Guid, StringBuilder> navigationSegments = new Dictionary<Guid, StringBuilder>();
 	private IDictionary<Guid, int> runningNavigations = new Dictionary<Guid, int>();
 
+	private IRegion? _initialRegion;
 
 	public void StartNavigation(INavigator navigator, IRegion region, NavigationRequest request)
 	{
+		_initialRegion = region;
```
kazo0, line 24: We would need to correlate the region to the specific NavigationRequest, no? We could have multiple navigation requests and we need to know which root region to use when EndNavigation is called.

src/Uno.Extensions.Navigation.UI/RouteNotifier.cs
```diff
@@ -53,7 +55,9 @@ public void EndNavigation(INavigator navigator, IRegion region, NavigationReques
 				Logger.LogTraceMessage($"Post-navigation (summary):\n{navigationSegments[id]}");
 			}
 			navigationSegments.Remove(id);
-			RouteChanged?.Invoke(this, new RouteChangedEventArgs(region.Root(), response?.Navigator));
+
+			var myRegion = _initialRegion ?? region;
```
kazo0, line 59: I'm assuming here is where we would need to resolve the correct `_initialRegion` based on the current `NavigationRequest` ID

PR #2843, opened 2025-07-07
ci: bump to latest ubuntu image

kazo0:


build/ci/stage-build-uitests-wasm.yml
```diff
@@ -4,7 +4,7 @@ jobs:
   container: unoplatform/wasm-build:3.0
 
   pool:
-    vmImage: 'ubuntu-24.04'
+    vmImage: 'ubuntu-latest'
```
ajpinedam, line 7: Isn't this "dangerous" because it will keep getting the latest version, making it hard to track which version is in use and the errors it brings?

kazo0: We seem to mostly use [ubuntu-latest](https://github.com/search?q=org%3Aunoplatform+ubuntu-latest&type=code) as opposed to [ubuntu-24.04](https://github.com/search?q=org%3Aunoplatform+ubuntu-24&type=code)

ajpinedam (approved):

Issue #2870, opened 2025-07-28
[Auth][OAuth2] Add oAuth2 as official Authentication option?


DevTKSS:
## What would you like to be added:

Authentication Extension that supports oAuth2 flow
- Add some Middleware(?) for IAuthenticationProvider to provide the capability of S253 challenge 
- Add capability for more Endpoint options: - #2874 
- Make public accessable, so we could add Properties/Funktionallity by inheritance?
  - the WebAuthenticationOptions / CustomAuthenticationOptions ...
  - Some WebAuthProvider (base?) class, not just the Interface or the BaseAuthenticationProvider, which does not already implement this Web specifics we would need for e.g. this szenario. By this, we would not have to care about how you call the Web Browser for example or other platform specifics

## Why is this needed:

I am looking for any way to implement oAuth2 into my Uno Applications.

## For which Platform:

- [x] iOS
- [x] Android
- [x] WebAssembly
- [x] WebAssembly renders for Xamarin.Forms
- [x] Windows
- [ ] Build tasks

## Anything else we need to know?

Found this Package on the WinAppSdk Repo here:
- https://github.com/microsoft/WindowsAppSDK/blob/main/specs/OAuth/OAuth2Manager.md

and on their issue here:
- https://github.com/microsoft/WindowsAppSDK/issues/441

someone posted this sample
- https://github.com/IlGalvo/WinUI3-OAuth2Manager-Sample
but seen that this is not supporting UWP, WinUI2 and so on also not really good, so I think that might not be a valid option.
- https://documentation.openiddict.com/integrations/operating-systems#supported-platforms

DevTKSS:
@jeromelaban dont know if Uno has this WinAppSdk issue on tracker list, but if not, would you consider adding it?

DevTKSS:
sample repro with web auth for a non protocol allowing api, using desktop target

[UnoWebAuthCodeFlowRepro.zip](https://github.com/user-attachments/files/21803694/UnoWebAuthCodeFlowRepro.zip)

added the delegates I used to put in my own project repo into a own service
- uno seems to expect us to do it the messy way with delegates so put it in the app.xaml.cs there, they are not connect able to the web auth builder as e.g. the Refresh would from the intellisense only provide 2 dictionarys but nothing else that would be needed!
[...]

Issue #2640, opened 2024-12-09
[Auth] Use custom IBrowser for OIDC/MSAL


kazo0:
Use something similar to https://github.com/IdentityModel/IdentityModel.OidcClient.Samples/blob/main/NetCoreConsoleClient/src/NetCoreConsoleClient/SystemBrowser.cs for a browser implementation that can support net-desktop targets

I believe @carldebilly has a modified version of this that can do away with the Kestrel dependency.

kazo0:
Poking @carldebilly / @nickrandolph for anymore info that we need for this one. 

Doing this will allow us to no longer rely on the `WebAuthenticationBroker`, correct?

carldebilly:
Yep. Actually I did something slightly different: instead of customizing the IBrowser which is specific to EntityModel.OidcClient (now renamed `Duende.IdentityModel.OidcClient`), I implemented something registering a custom WebAuthenticationBroker into Uno (using api extensibility mechanisms built into Uno) and used the [AutoRedirectUri](https://github.com/unoplatform/uno.extensions/blob/main/src/Uno.Extensions.Authentication.Oidc/OidcAuthenticationSettings.cs#L10) feature to correctly resolve the uri from the browser.

The missing piece is the custom WAB, which is using [my "new" http server](https://github.com/carldebilly/Yllibed.HttpServer) which is extremely light and has no dependencies. Far easier to use than Kestrel.

DevTKSS:
Adding link to discord server messages including problems with the above shown code
- https://discord.com/channels/1182775715242967050/1193995186951426048/1401619641171837018

- [ ] the properties in the top if-statements coming from the `Windows.Security.Authentication.Web` are beeing dropped from implemenation (`UNO001` Not Implement`) from `Uno.Sdk="6.0.146"` upwards (tested also with the latest dev build of the sdk today using `global.json` with the latest pre release nuget version and `allowPrerelease = true`)
- [ ] `IWebAuthenticationBrokerProvider` interface or more the `Uno.AuthenticationBroker` Namespace is not longer known in Apps using the `Uno App` vs2022 vsix template, **But** using a `Uno Library` project that's resolving fine.
[...]

PR #2890, opened 2025-08-10
feat(WebAuthenticationProvider): Allow providing IWebAuthenticationBrokerProvider via DI parameters

DevTKSS:

## PR Type

- We could register that Interface in DI conditionally, as the WebAuthenticationProvider is no public type or any of the Web Auth parts, its not using an Interface then a class itself.
- While Uno does have this Broker and Interface definition it would be good to have a chance to use this feature 👍 
- Missing oAuth Code Challenge callback function

## What is the new behavior?

- If we would register an Implementation (e.g. via ApiExtension) of this Interface, the DI constructor would resolve IWebAuthenticationBroker as not null
- Optional Argument so existing Code will not break, while Users of Web Auth could provide ApiExtensions of this to be used.
- Applyed Conditional `#if !WINDOWS` because of otherwhise this Interface gets linted with CS0246. see issue on the unoplatform/uno repo for more information about this:
  - https://github.com/unoplatform/uno/issues/21237

- Should enable also to use any future Implementations of this Interface to be used like I am trying to acomplish right now just with HttpListener or maybe Refit for enabling https instead of just http (tpc Listener seems to not be able to use https?):
  - https://github.com/unoplatform/uno.extensions/issues/2640

- added a comment where the oAuth code challenge function should be made availeble as minimalistic approach/or hopefully solution (?) for adding custom action for this without having to copy and re-write the whole WebAuthentication classes and so on, which uno has already!

## PR Checklist

  - Not possible for me to build it locally, as its having that unknown numeric targetframework error

## Other information

As I am not sure until now how the signature of this function should look like as none of the WebConfiguration Functions does have proper xml docs at least or speaking variable names beside the Name of the function like Login or logout, I only added that comment, but hopefully someone of the more expirienced uno devs could install a little additional available `AsyncFunc` right there, so I could "just" add another delegate like the PrepareLoginUri function and do the missing oAuth code flow between Authorization and the point where the WebAuthProvider could proceed with the Tokens actions.

As you can lookup in the following link, the stage of codeflow at the commented point would not include any parameter with `access_token` or `refresh_token` so that could not work out like right now setup to process login properly.
- [Link to the official oAuth standard document](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2)

if this callback could be added, this could be a usable workaround until a full oauth feature would be added for:
- https://github.com/unoplatform/uno.extensions/issues/2870

### DIFF
--- a/src/Uno.Extensions.Authentication.UI/Web/WebAuthenticationProvider.cs
+++ b/src/Uno.Extensions.Authentication.UI/Web/WebAuthenticationProvider.cs
@@ -1,5 +1,7 @@
 ﻿using Microsoft.Extensions.Options;
-
+#if !WINDOWS
+using Uno.AuthenticationBroker;
+#endif
 namespace Uno.Extensions.Authentication.Web;
 
 internal record WebAuthenticationProvider
@@ -8,7 +10,10 @@ internal record WebAuthenticationProvider
 	IOptionsSnapshot<WebConfiguration> Configuration,
 	IServiceProvider Services,
 	ITokenCache Tokens
-) : BaseAuthenticationProvider(ProviderLogger, DefaultName, Tokens)
+#if !WINDOWS
+	,IWebAuthenticationBrokerProvider? WebAuthBroker = null
+#endif
+	) : BaseAuthenticationProvider(ProviderLogger, DefaultName, Tokens)
 {
 	private const string OAuthRedirectUriParameter = "redirect_uri";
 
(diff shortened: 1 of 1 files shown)


DevTKSS:
removed Todo for callback would be able to work with if this PR and the #2893 would get merged, so we could provide the WebAuthBrokerProvider Implementation and the `OtherTokenKeys` Dictionary from appsettings.json would get even considered. at this time, thats not used at all and the accessor Level is not allowing also that we would get the WebConfiguration via IOptions, as mentioned on the other PR.

@kazo0 would you see the option that this could get reviewed? then I could just put an Implementation in the DI container on non Windows targets like Desktop and could go on with my Projects. e.g. my approach would be HttpListener with OpenBrowser like we seen on #2640 but without the tcpListener using approach because my API I need does require other then what this Server setup can do because its not allowing to tell that I would like to have localhost and not `127.0.0.0` for example or to use https and so on. to match the interfaces it does introduce is absolutly horror and no quick doing, so I would rather just match the interface of this here... together with IOptions I would expect it to work... otherwise I can only drop webauth extension until Uno has a implementation working for desktop and creat a IAuthenticationService which would be not very clean against this easy doing here...

PR #2892, opened 2025-08-12
feat(oAuth): Add OAuth needed Configuration Properties and callback delegate for Code Exchange

### LINKED ISSUE #2874: [Authentication][Options][Configuration] Missing Properties TokenEndpoint and UserInfoEndpoint for oidcAuth & WebAuthenticationOptions
## Current behavior

If using Windows target with (assuming) oAuth2Manager we currently seem not to have the option of providing an:
- Authentication Endpoint (Authority != Authentication for oAuth2)
- TokenEndpoint
- UserInfoEndpoint

- Windows target seems to not provide any https:// prefixed url ? => not working for this api

- Desktop Target Breaks on "OnLaunched" with "StackOverflowException" (no more details available)
- Web Auth also does not have any of those Properties we could set via options/configuration

## Expected behavior

these Options Members should be:
- available to be set through configuration Options!
- set-able through just like `AddOidc(configure => configure.Authority(` without beeing required to use the IdentityModel OidcClientOptions in case this package would might get replaced in any future, this direct access through the Uno Extension provided Methods would reduce the breaking changes then
- Should be also available via any non oidc auth, like web auth would also do, 

### Theoretically Workaround

set through the ProviderInformation Property like this

Image
but still does not work using oidc #2872 on wasm as told

clicking on Login does NOTHING, 

- Look into the documentation -> not told to be available
- Look into the src -> Not checked for

## Environment

Windows 11 Pro

Nuget Package (s):

Package Version(s):

Affected platform(s):

- [x] all Platforms
- [x] Build tasks

Visual Studio:

- [x] Visual Studio 2022 17.14.9

Relevant plugins:

- [ ] Resharper (version: )

not missunderstand as duplicate of the oidc not working at all on wasm (tryed this on android, desktop, windows, nothing works)
this is about that those Options or Settings are missing on the appsettings provided ones to be set for if we do not have a oidc then a oAuth2 api to make calls to, so you do not have any discovery endpoint.

tryed with uno sdk 6.0.146 and 6.1.23

DevTKSS:
TODO: Missing integration in WebAuthenticationProvider and the Delegate should get checked which of the parameters are required and which return Type this should have

GitHub Issue (If applicable): closes #2874 
(possibly also more)

## PR Type

- Feature

- The e.g. TokenEndpoint is missing, but required if the api does not support a discovery Endpoint and we want the WebAuth to integrate this oAuth flow
- no Authorization Code Exchange Delegate

## What is the new behavior?

- Added the Properties I think this will need in the end (refering to asp net core oAuth Properties) so we not need to load multiple Settings/Options just because the main ones are incomplete
- Added an Delegate for Code Exchange (Help needed here)

## PR Checklist

- [ ] Tested code with current [supported SDKs](../README.md#supported) *not possible, I could not build extensions locally even if this would be ready up to this state*
- [ ] Contains **NO** breaking changes 🤔 *not sure?*

## Other information

I would need somone helping on finishing the Code Exchange callback up please.
Along the guide for oAuth of the Api I need to request, this Service of mine should work with oAuth generally.
I used Refit and put the Delegates all in one class to get the DI options for it [In my Project](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/blob/master/src/DevTKSS.MyManufacturerERP/Infrastructure/Services/EtsyOAuthAuthenticationService.cs)

Telling you about this to allow you checking if you migth find some things like the defaults and oAuth2Utility (would defintly make sense in my opinion) that could be added as resource to be used from all that might need those functionallities in the future like me. this is so far NOT part of this PR contents dont worry ;) but let me know if you would like some of this to be added 👍 

- [this here are my oAuthOptions it is reffering to](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/blob/master/src/DevTKSS.MyManufacturerERP/Models/OAuthOptions.cs)
- [and here you can see what hell of a mess its doing to be needed to create multiple settings for one api](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/blob/master/src/DevTKSS.MyManufacturerERP/appsettings.development.json) so thats also one of the reasons it would make sense to feed Auth and Http from the same Instance by just adding the minimum...
- [Defined constant Defaults like the asp net core oAuth so this takes over faults by typos](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/tree/master/src/DevTKSS.MyManufacturerERP/Infrastructure/Defaults)
- [oAuth2Utilities](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/blob/master/src/DevTKSS.MyManufacturerERP/Infrastructure/OAuth2Utilitys.cs)

Thanks in advance for checking on this!

### DIFF
--- a/build/ci/cspell.json
+++ b/build/ci/cspell.json
@@ -55,6 +55,8 @@
 		"Omnisharp",
 		"overscroll",
 		"Packt",
+		"Pkce",
+		"PKCE",
 		"parameterless",
 		"pickable",
 		"Pluralsight",
--- a/src/Uno.Extensions.Authentication.UI/Web/WebAuthenticationSettings.cs
+++ b/src/Uno.Extensions.Authentication.UI/Web/WebAuthenticationSettings.cs
@@ -11,7 +11,9 @@ internal record WebAuthenticationSettings
 	public string? LoginCallbackUri { get; init; }
 
 	public AsyncFunc<IServiceProvider, ITokenCache, IDictionary<string, string>?, string?, string>? PrepareLoginCallbackUri { get; init; }
-
+	public AsyncFunc<IServiceProvider, ITokenCache, IDictionary<string, string>?, string?, string?> ExchangeCode { get; init; }
+	public string? TokenEndpoint { get; init; }
+	public bool UsePkce { get; init; } = false;
 	public string AccessTokenKey { get; init; } = "access_token";
 	public string RefreshTokenKey { get; init; } = "refresh_token";
 
--- a/src/Uno.Extensions.Authentication.UI/Web/WebConfiguration.cs
+++ b/src/Uno.Extensions.Authentication.UI/Web/WebConfiguration.cs
@@ -5,6 +5,9 @@ internal record WebConfiguration
 	public bool PrefersEphemeralWebBrowserSession { get; init; }
 	public string? LoginStartUri { get; init; }
 	public string? LoginCallbackUri { get; init; }
+	public string? TokenEndpoint { get; init; }
+	public string? UserInfoEndpoint { get; init; }
+	public bool UsePkce { get; init; } = false;
 	public string? AccessTokenKey { get; init; }
 	public string? RefreshTokenKey { get; init; }
 	public string? IdTokenKey { get; init; }


DevTKSS:
the diff seems to be bugged on the WinUIEx_WebAuthenticator filed as VS2022 issue and it seems to be caused by unpurposely lineending changes. diff when disabling those in the vs ide:

image

so that change in this file is only this line on this commit.
same happened annoyingly on the WebConfiguration file. maybe I need to drop those commits, not sure how to fix those issues caused by the ide...
image

image

(removed callback which can get resolved by
- #2893 
which introduces recognizing the OtherTokenKeys where we could add the parameters we need additionally like `error` `error_description` `state` and `code` from the Access Grant response of the Authorization Code flow
@kazo0 FYI

DevTKSS:
in my opinion, the oAuth normal `client_id` parameter should also be known from the WinUiEx class since the `id_token` is something JWT only reading in the internet.

Issue #2876, opened 2025-07-31
[Auth] Update/Create reference samples for all Auth


kazo0:
We need to audit, update, and/or create new reference samples for OIDC, MSAL, Web Auth, and Custom Auth.

The current state of Auth is unclear at this point and we need a reliable way of testing the latest bits of Auth Extensions. These should also include samples of authenticated HTTP calls made using the HTTP Extensions through both Refit and Kiota.

Issue #2875, opened 2025-07-31
[Auth][OIDC] Transition plan from IdentityModel/Duende


kazo0:
We need to determine if we can move away from IdentityModel.OidcClient as the underlying implementation for the `OidcAuthenticationProvider`. This is not deprecated and replaced with [Duende IdentityModel](https://docs.duendesoftware.com/identitymodel) and it isn't clear whether we should be moving to this as this may be a paid service.

Need to discuss options with @carldebilly / @nickrandolph

DevTKSS:
I think OpenIdDict could be a at least free option #2870

DevTKSS:
@kazo0 a point to consider for this to be made change I would like to add:

- Could we have the Xamarin [...] Browser Package updated?

The reason(s) for this:

- If we need to use them to for example embedd Google advertisements banners in our apps (you could search the discord server for the user that did this) limiting the version to net 8

Would suggest the OpenIdDict package for oidc future plans to consider:

- the OpenIdDict.Client.SystemIntegration Nuget we would need to use in our app does require that higher Xamarin packages version

- seen this OpenIdDict package also enable OAuth2 🤔 just on Windows target in that sample but generally the package seems to work on other targets too.

- Only limitation of it could potentially be the UWP/ Win32 Desktop target but this may be solve-able with a proper Desktop target support for the `WebAuthenticationBrokerProvider`.

- Also good for this (used by current web and oidc auth extensions) would be to maybe allowing to DI inject the Interface by adding another ctor on the `WebAuthenticationProvider` thats encapsulating the platform specific implementation for example also used by our current Web and oidc auth provider and by that have the uno ApiExtension Attribute problem surrounded, which makes the build fail if we create this implementation but have SOC applied.

DevTKSS:
Should be fixed alongside with the transition:
https://github.com/unoplatform/uno.extensions/issues/2895
Because otherwise its not even resolving the auth provider

PR #2893, opened 2025-08-16
BREAKING CHANGE: TokenCacheExtensions: Match the used key namings to the TokenCache defaults and standarized official namings

DevTKSS:

## PR Type

- Documentation content changes
- Feature
- Other... Please describe: Future Bug Prevention!
- Bugfix

- Assuming somone would use the TokenCacheExtension defined Extensions, which do not allow for those Keys to provide an alternative, the tokens would not be able to receive using that constants (e.g. I would like to use `access_token` and `refresh_token` but that would be failing if I would not directly set this in the options I am not told about as regular user(!) the WebAuthenticationProvider would always use the Constants with the other casing...
- The `OtherTokenKeys` are not evaluated at all, the provider does not care about them and the WebConfiguration type couldnt either be gotten through `IOptions<WebConfiguration>` so no chance to get them again if used. But we are neither told that they exist at all from the docs. is that because of this not implementation of it?

### Bugfix / -Prevention

- The keys used by the `TokenCacheExtensions` do now match the keys used in the original TokenCache
This will ensure stability at our Apps runtime and prevent the issue that users of the Extension Methods would not be able to resolve their before stored Tokens by the url parameter names.

### Features:

- added optional parameters to the SaveTokensAsync method that will allow to use it with own Keys
- added Extensions for the other standarized Keys alongside with the relevant const string representations
- added to each of the seperate existing methods for e.g. AccessTokenAsync the key name as optional parameter

  By allowing the key as optional parameter and those new extensions, we will gain more flexibility for the user of this.
  This will be especially usefull for cases like oAuth where we do not have a id_token then e.g. a user_id that would be important to store instead of that and would represent in this scenario the equivalent to the id_token of jwt.

- Added Dictionary Extension so in case that we would not have this key in the dictionary so far, we would not run into issues if attempting to add it.

  This would also be usefull if the `OtherTokenKeys` would be evaluated to be used at all in the web auth

- Added Fluent Builder for the `TokenCacheOptions` and implementation in the Hostbuilder Extensions just like the oidc extension already does, but adjusted to its nested structure in `WebConfiguration` and `WebAuthenticationSettings`

## PR Checklist

  - so far I dont know how exactly to do this the docs are attempting to tell for the uno.ui repo but seems like the Web project does miss packages?
  Not sure if `RebuildApp: 44 succeeded, 3 failed, 0 skipped` Solution Build result does satisfy your question about "Tested code with current supported SDKs" but here is the Build solution output:

  [Output-Build.txt](https://github.com/user-attachments/files/21816981/Output-Build.txt)

  not so far, can do if the changes are okay with the team.
  - failing to add tests for e.g. the Dictionary Extension in the core extensions project, but thats not possible and causing a circular reference
  - Build of the Solution (without filter) does fail on runtimeTests?
- [x] Associated with an issue (GitHub or internal) partially causing some of my issues to oAuth but maybe not closing them completly

If this PR contains a breaking change, please describe the impact and migration path for existing applications below.
     Please note that breaking changes are likely to be rejected

- AuthenticationBuilderExtensions would be required to be adjusted for the moved Options to the `TokenCacheOptions` Type
- the seperated type can be inherited from like the `EndpointOptions` not just from Extensions own classes then also from the using uno app, so they can be gotten throught `IOptions<TokenCacheOptions>` using e.g. the Path `Web:TokenOptions` or also its DefaultName which is defined as const, so we have no need to type it our own then can use that const instead for NamedOptions.
- We are now able to source the Parameters from the `OtherTokenKeys` provided in there as the `WebAuthenticationProvider` does now have a foreach loop set up, which is extracting each of the Parameters by its Value provided urlKey.

## Other information

- needs help to implement tests for this
- not really seeing the way to "build it locally" from the docs. Without target framework override, so building everything, doesnt that satisfy this point? not sure...

### DIFF
--- a/src/Uno.Extensions.Authentication.UI/HostBuilderExtensions.cs
+++ b/src/Uno.Extensions.Authentication.UI/HostBuilderExtensions.cs
@@ -37,10 +37,7 @@ public static class HostBuilderExtensions
 		hostBuilder
 			.UseConfiguration(configure: configBuilder =>
 					configBuilder
-						.Section<WebConfiguration>(name)
-						.Section<TokenCacheOptions>(string.Join(':', name, TokenCacheOptions.DefaultSectionName))
-				);
-
+						.Section<WebConfiguration>(name));
 
 		var authBuilder = builder.AsBuilder<WebAuthenticationBuilder>();
 
@@ -90,9 +87,8 @@ public static class HostBuilderExtensions
 		hostBuilder
 			.UseConfiguration(configure: configBuilder =>
 					configBuilder
-						.Section<WebConfiguration>(name)
-						.Section<TokenCacheOptions>(string.Join(':', name, TokenCacheOptions.DefaultSectionName))
-				);
+						.Section<WebConfiguration>(name));
+		
 		var authBuilder = builder.AsBuilder<WebAuthenticationBuilder<TService>>();
 
 		configure?.Invoke(authBuilder);


DevTKSS:
Related issues from the team for the plans with Oidc to lookup:
#2875 #2876

PR #2907, opened 2025-10-08
chore: avoid Assembly.GetCallingAssembly()

jonpryor:
Context: https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Assembly.NativeAot.cs#L15-L26

`System.Reflection.Assembly` has a few constraints when running
within a NativeAOT environment:

  * `Assembly.GetEntryAssembly()` works.
     Note, though, that this method returns `null` on Android when
     using MonoVM and CoreCLR.

  * `Assembly.GetExecutingAssembly()` always throws `NotImplemented.ByDesign`.

  * `Assembly.GetCallingAssembly()` throws `PlatformNotSupportedException`
    *by default*.

The exception thrown by `Assembly.GetCallingAssembly()` can be avoided
by adding a `Switch.System.Reflection.Assembly.SimulatedCallingAssembly`
"runtime switch" to the app:

	<ItemGroup>
	  <RuntimeHostConfigurationOption
	      Include="Switch.System.Reflection.Assembly.SimulatedCallingAssembly"
	      Value="true"
	      Trim="true"
	  />
	</ItemGroup>

However, this is *not* the default, and is undocumented.  (UTSL!)

For future NativeAOT support, add a `PlatformHelper.GetAppAssembly()`
method and update all existing `Assembly.Get*Assembly()` methods to
instead use `PlatformHelper.GetAppAssembly()`.

`PlatformHelper.GetAppAssembly()` uses `Assembly.GetCallingAssembly()`;
if that fails, `Assembly.GetEntryAssembly()` is returned instead.
If a NativeAOT-like environment is detected, only
`Assembly.GetEntryAssembly()` is used.

Enforce use of `PlatformHelper.GetAppAssembly()` by using the
[Microsoft.CodeAnalysis.BannedApiAnalyzers][0] NuGet package to "ban"
use of those `Assembly.Get*Assembly()` methods; attempts to use them
will result in an RS0030 error.  (This can be ignored if necessary
via `#pragma warning disable RS0030`.)

[0]: https://www.nuget.org/packages/Microsoft.CodeAnalysis.BannedApiAnalyzers/

## PR Type

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Extensions.Core/PlatformHelper.cs
+++ b/src/Uno.Extensions.Core/PlatformHelper.cs
@@ -60,29 +60,62 @@ public static class PlatformHelper
 		}
 	}
 
-	internal static void SetAppAssembly(Assembly assembly) => _appAssembly = assembly;
-
 	/// <summary>
-	///   Attempts to obtain the "App" <see cref="T:Assembly"/>.
+	///  Set the value returned by <see cref="GetAppAssembly"/>.
+	/// </summary>
+	/// <param name="assembly">
+	///  The <see cref="Assembly"/> that subsequent calls to <see cref="GetAppAssembly" /> should return.
+	/// </param>
+	/// <remarks>
+	///   <para>If <paramref name="assembly"/> is <see langword="null" />, then
+	///   <see cref="GetAppAssembly"/> will follow its default algorithm.
+	///   </para>
+	/// </remarks>
+	public static void SetAppAssembly(Assembly? assembly) => _appAssembly = assembly;
+
+#pragma warning disable RS0030
+	/// <summary>
+	///   Attempts to obtain the "App" <see cref="Assembly"/>.
 	/// </summary>
 	/// <returns>
-	///   An <see cref="T:Assembly"/> if the "Application" assembly can be found; otherwise, <see langword="null"/>.
+	///   An <see cref="Assembly"/> if the "Application" assembly can be found; otherwise, <see langword="null"/>.
 	/// </returns>
 	/// <remarks>
-	///   <para>The complication is NativeAOT, within which:</para>
-	///   <list type="bullet">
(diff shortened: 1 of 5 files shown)


src/Uno.Extensions.Core/PlatformHelper.cs
```diff
@@ -56,6 +60,47 @@ public static bool IsAppPackaged
 		}
 	}
 
+	internal static void SetAppAssembly(Assembly assembly) => _appAssembly = assembly;
```
jonpryor: We should probably make this and `GetAppAssembly()` public, as we may need to use its functionality in e.g. https://github.com/unoplatform/ShowMeTheXAML/blob/aae225c269c7fa267d26738b615c61a769f55399/ShowMeTheXAML/XamlDisplay.cs#L67.

We may also want to make the parameter nullable, so that `PlatformHelper.SetAppAssembly(null)` is permitted.  This would mean "use the default value next time."

testing/TestHarness/TestHarness/BaseHostInitialization.cs
```diff
@@ -7,7 +7,7 @@ public abstract class BaseHostInitialization : IHostInitialization
 	public virtual IHost InitializeHost(Window window)
 	{
 		var host = UnoHost
-				.CreateDefaultBuilder()
+				.CreateDefaultBuilder(Assembly.GetEntryAssembly())
```
jonpryor: While this is for testing, this won't work on Android with MonoVM and CoreCLR, as `Assembly.GetEntryAssembly()` returns `null` there.

Perhaps this code won't ever run on Android?  But if it *does*, then will this work as expected/intended?

nickrandolph (approved):

PR #2913, opened 2025-10-20
fix(ThemeService): Only save theme on explicit user action

KL2400040448:
Closes #2842

## PR Type

- Bugfix

When the application is set to follow the system theme, any change in the system's theme (e.g., from Light to Dark) causes the `ThemeService` to save the resolved theme ("Dark") to the settings file. This locks the application to that specific theme and prevents it from following any subsequent system theme changes.

## What is the new behavior?

This change moves the theme-saving logic from the internal theme-changed handler (`InternalSetThemeOnUIThread`) to the public `SetThemeAsync` method.

This ensures that the theme preference is only saved to settings when a user explicitly calls `SetThemeAsync` with a specific theme (Light or Dark). The application now correctly follows the system theme by default without writing to the settings file, which resolves the original issue.

## PR Checklist

- [ ] Docs have been added/updated which fit documentation template. (for bug fixes / features)
- [ ] Unit Tests and/or UI Tests for the changes have been added (for bug fixes / features) (if applicable)
- [ ] Wasm UI Tests are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [ ] Updated the Release Notes

### DIFF
--- a/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
+++ b/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
@@ -99,8 +99,15 @@ internal class ThemeService : IThemeService, IDisposable
 			throw new NullReferenceException($"Theme service not initialized, {nameof(InitializeAsync)} needs to complete before SetThemeAsync can be called");
 		}
 
-		// Make sure initialization completes before attempting to set new theme
-		await _initialization.Task;
+		if (theme != AppTheme.System)
+                {
+                         await _settings.SetAsync(nameof(CurrentTheme), theme.ToString());
+                }
+
+                if (theme != AppTheme.System)
+                {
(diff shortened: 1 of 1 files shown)


KL2400040448:
review

KL2400040448:
merge approve

src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
```diff
@@ -151,7 +160,6 @@ private void SaveDesiredTheme(AppTheme theme)
 	{
 		try
 		{
-			_settings.Set(CurrentThemeSettingsKey, theme.ToString());
```
kazo0: What's the reason for removing this line and leaving an empty `try` block

PR #3062, opened 2026-04-23
fix: Prevent ThemeService from leaking theme across XamlRoot boundary

### LINKED ISSUE #3061: ThemeService leaks theme across XamlRoot boundary in ALC-hosted scenarios
## Description

`ThemeService.InternalSetThemeOnUIThread` resolves the target element for `RequestedTheme` via `RootElement?.XamlRoot?.Content`. When the `ThemeService`'s `RootElement` is hosted inside another application's visual tree (e.g., via `AlcContentHost` for AssemblyLoadContext isolation), both the host and guest share the **same `XamlRoot`**. This causes `XamlRoot.Content` to resolve to the **host's root element** rather than the guest's own root, leaking the theme change to the entire host application.

## Reproduction

1. Host an Uno app inside another Uno app using ALC isolation (`AlcContentHost`). The guest app must use `UseToolkitNavigation()` or `UseToolkit()` (which registers `ScopedThemeService`).
2. Change the `RequestedTheme` on an ancestor element of the guest app's content (e.g., toggle between Dark and Light).
3. The guest app's `ThemeService` reacts to `ActualThemeChanged` on its `RootElement` and sets `RequestedTheme` on `XamlRoot.Content` — which is the **host's** root.
4. The host application's theme changes unexpectedly.

### Visual tree structure

```
Host's Root  ← XamlRoot.Content resolves HERE
  ├── Host Toolbar
  ├── Host Sidebar
  └── Container
        └── AlcContentHost
              └── Guest App Root  ← ThemeService.RootElement
```

Both live in the same `XamlRoot`. When `ThemeService` navigates `RootElement.XamlRoot.Content`, it escapes the guest boundary and targets the host's root.

### Additional amplifying factors

- **Shared `ISettings`**: Both host and guest `ThemeService` instances share `ApplicationData.Current.LocalSettings` via the same `Settings` implementation. The host saves `"Dark"` to the `"CurrentTheme"` key; the guest reads it, causing `ElementThemeChanged` to bypass the `AppTheme.System` early-return and enter the `InternalSetThemeAsync` code path that triggers the leak.

- **Stale `ActualTheme` during theme walk**: `RaiseActualThemeChanged()` fires before `SetTheme()` in `NotifyThemeChangedCore`, so `sender.ActualTheme` returns the previous theme value. This creates a one-toggle-delayed leak pattern where the host's theme changes on the second toggle rather than the first.

## Expected behavior

`ThemeService` should set `RequestedTheme` on `RootElement` directly, keeping the theme change scoped to its own subtree — not navigate through `XamlRoot.Content` which can escape the content boundary.

## Proposed fix

Change `InternalSetThemeOnUIThread` line:

```diff
-var rootElement = RootElement?.XamlRoot?.Content as FrameworkElement;
+var rootElement = RootElement as FrameworkElement;
```

For normal (non-ALC) apps, `RootElement` (`window.Content`) and `XamlRoot.Content` (`WindowChrome.Content`) are the same element — no behavioral change. For ALC-hosted apps, this keeps the theme change within the guest's subtree.

MartinZikmund:
## Summary

- `ThemeService.InternalSetThemeOnUIThread` navigated `RootElement?.XamlRoot?.Content` to find the target for `RequestedTheme`. When the `RootElement` is hosted inside another app's visual tree (ALC isolation via `AlcContentHost`), both share the same `XamlRoot`, so `XamlRoot.Content` resolves to the **host's root** — leaking the theme change to the entire host.
- Changed to use `RootElement` directly. For normal apps, `RootElement` (= `window.Content`) and `XamlRoot.Content` (= `WindowChrome.Content`) are the same element — no behavioral change. For ALC-hosted apps, this keeps `RequestedTheme` scoped to the guest's own subtree.

### The visual tree in ALC hosting

```
Host Root  ← XamlRoot.Content resolves HERE
  ├── Host UI elements
  └── AlcContentHost
        └── Guest App Root  ← ThemeService.RootElement
```

Both live in the same `XamlRoot`. `ThemeService` was escaping its boundary by navigating `RootElement.XamlRoot.Content` → host's root.

### Contributing factors

1. **Shared `ISettings`**: Host and guest `ThemeService` instances both use `ApplicationData.Current.LocalSettings` via the `Settings` class. Host saves `"Dark"` to `"CurrentTheme"`, guest reads it — bypassing the `AppTheme.System` early-return in `ElementThemeChanged` and entering the `InternalSetThemeAsync` path.

2. **Stale `ActualTheme` during theme walk**: In Uno Platform's `NotifyThemeChangedCore`, `RaiseActualThemeChanged()` fires before `SetTheme()`, so `sender.ActualTheme` returns the **previous** value. This creates a one-toggle-delayed leak pattern.

### The fix

```diff
-var rootElement = RootElement?.XamlRoot?.Content as FrameworkElement;
+var rootElement = RootElement as FrameworkElement;
```

## Test plan

- [ ] Verify normal (non-ALC) theme switching still works — `RootElement` and `XamlRoot.Content` are the same element, so no behavioral difference
- [ ] Verify ALC-hosted app theme toggle does not leak to the host app
- [ ] Verify `SetThemeAsync(Dark/Light/System)` works correctly in both normal and ALC scenarios

Closes #3061

### DIFF
--- a/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
+++ b/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
@@ -133,7 +133,7 @@ internal class ThemeService : IThemeService, IDisposable
 	private bool InternalSetThemeOnUIThread(AppTheme theme)
 	{
 		var existingIsDark = IsDark;
-		var rootElement = RootElement?.XamlRoot?.Content as FrameworkElement;
+		var rootElement = RootElement as FrameworkElement;
 
 		if (rootElement is null)
 		{


src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
```diff
@@ -133,7 +133,7 @@ private async Task<bool> InternalSetThemeAsync(AppTheme theme)
 	private bool InternalSetThemeOnUIThread(AppTheme theme)
 	{
 		var existingIsDark = IsDark;
-		var rootElement = RootElement?.XamlRoot?.Content as FrameworkElement;
```
jeromelaban, line 136: let's explain why we're not using the inner root, for future reference.

jeromelaban (approved):

morning4coffe-dev (approved):
Reviewed the one-line change at `9e611554b5d5c4691c6f6f964821d095483d1de1` and the surrounding initialization, theme observation, dispatch and disposal paths. Targeting the service's own root rather than escaping through XamlRoot.Content is the correct scope fix.

I ran the exact base/head ThemeService implementations in a native Windows App SDK 2.4.0/.NET 10 app with real host, guest and sibling elements sharing one XamlRoot. The base's guest Dark request darkened the host and sibling; the head kept them Light while applying Dark only to the guest. Light/System requests, inherited parent changes and normal Window-based Dark/Light switching also behaved as expected. The Dark request exercised the native UI dispatcher from a worker thread.

One observable contract detail: detached FrameworkElements now finish initialization and retain the requested theme before loading. Please update IThemeService's old 'Fails if no XamlRoot found' wording and add the explanation requested in the existing review thread; I have not resolved another reviewer's thread.

Scope: seven observations per implementation, exact service/helper source, in-memory settings and a native dispatcher adapter. No secondary ALC, shared ApplicationData persistence, Uno Skia/mobile/WASM or OS-theme-change execution is claimed.

PR #3063, opened 2026-04-23
Test/tabbar hotreload tests

vatsashah45:
GitHub Issue (If applicable): https://github.com/unoplatform/private/issues/926
https://github.com/unoplatform/uno.extensions/issues/2912
https://github.com/unoplatform/uno.extensions/issues/2971
https://github.com/unoplatform/uno.extensions/issues/3075
https://github.com/unoplatform/uno.extensions/issues/3086
https://github.com/unoplatform/uno.extensions/issues/3088

## PR Type

What kind of change does this PR introduce?

- Bugfix
- Feature
- Code style update (formatting)
- Refactoring (no functional changes, no api changes)

## What is the new behavior?

This pull request adds new test infrastructure and supporting classes for hot-reload scenarios involving TabBar navigation in the Uno.Extensions.Navigation UI test suite. The changes introduce dedicated test targets, view models, and page classes to enable isolated and robust hot-reload testing for TabBar components, including scenarios with two and three tabs. Additionally, the test project now references the Uno.Extensions.Navigation.Toolkit to support these tests.

**New TabBar Hot-Reload Test Infrastructure:**

* Added new test classes for TabBar hot-reload scenarios:
  - `HotReloadTabBarTarget` and `HotReloadTabGateTarget` provide isolated hot-reload targets to prevent delta cache interference between tests.
  - `HotReloadTabBarVm` view model exposes the value used in tab content, delegating to the hot-reload target.
  - `HotReloadTabContentPage`, `HotReloadTabBarPage`, and `HotReloadTabBarThreeTabPage` provide XAML page structures for two-tab and three-tab TabBar hot-reload tests, and for displaying tab content.

**Project and Test Setup:**

* Updated the test project to reference `Uno.Extensions.Navigation.Toolkit` to enable toolkit navigation features required by the new TabBar tests.
* Registered the new `Given_TabBarHotReload` test class in the test suite initialization to ensure the TabBar hot-reload tests are executed in debug builds.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tested code with current [supported SDKs](../README.md#supported)
- [ ] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md). (for bug fixes / features) 
- [ ] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
[...]

PR #3066, opened 2026-04-27
fix: Re-evaluate route map after C# hot-reload metadata update

vatsashah45:
GitHub Issue (If applicable):

Related to https://github.com/unoplatform/uno.extensions/pull/3063

Related to https://github.com/unoplatform/private/issues/926

## PR Type

- Bugfix

## What is the new behavior?

Add NavigationRouteUpdateHandler as a standard .NET MetadataUpdateHandler that re-invokes the viewRouteBuilder delegate registered via UseNavigation and rebuilds RouteResolver mappings when C# hot-reload applies a delta.

- Registry<T>: add Clear() to allow re-invocation of route builder
- RouteResolver: add Rebuild() to re-process registries into flat lookups
- ServiceCollectionExtensions: store delegate/registry/resolver references for the hot-reload handler to access

## PR Checklist

## Other information

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
+++ b/src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
@@ -5,6 +5,19 @@ using System.Reflection.Metadata;
 namespace Uno.Extensions.Navigation.UI;
 
 /// <summary>
+/// Holds the references needed to refresh navigation routes after a C# hot-reload.
+/// One instance exists per navigation engine (<see cref="ServiceCollectionExtensions.AddNavigation"/>).
+/// Registered/unregistered by <see cref="NavigationHostedService"/> to avoid leaking.
+/// </summary>
+internal sealed class NavigationRouteContext
+{
+	public required Action<IViewRegistry, IRouteRegistry> RouteBuilder { get; init; }
+	public required IViewRegistry Views { get; init; }
+	public required IRouteRegistry Routes { get; init; }
+	public RouteResolver? Resolver { get; set; }
+}
+
+/// <summary>
 /// C# hot-reload handler that re-invokes the <c>viewRouteBuilder</c> delegate
 /// registered via <see cref="ServiceCollectionExtensions.AddNavigation"/> and
 /// rebuilds <see cref="RouteResolver"/> mappings so that newly added routes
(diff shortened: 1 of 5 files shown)


src/Uno.Extensions.Navigation.UI/ServiceCollectionExtensions.cs
```diff
@@ -21,6 +22,11 @@ public static IServiceCollection AddNavigation(
 		var routes = createRouteRegistry?.Invoke(services) ?? new RouteRegistry(services);
 		routeBuilder?.Invoke(views, routes);
 
+		// Store references for C# hot-reload route refresh.
+		NavigationRouteUpdateHandler.RouteBuilder = routeBuilder;
+		NavigationRouteUpdateHandler.Views = views;
+		NavigationRouteUpdateHandler.Routes = routes;
```
dr1rrb: This will leak the navigation engine and assume that we have only one instance of nav engine per app. Even if usually we have only one per app, this is still possible to have multiple instances!

We should rely on the `IHostedService` (I see a `NavigationHostedService` on line 34) to register on `StartAsync` and un-register on `StopAsync` a reference to the service (in a list) in the `NavigationRouteUpdateHandler`.

src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+			viewRegistry.Clear();
+		}
+
+		if (routes is Registry<RouteMap> routeRegistry)
+		{
+			routeRegistry.Clear();
+		}
+
+		// Re-invoke the (potentially updated) delegate.
+		builder.Invoke(views, routes);
```
dr1rrb: We can try with this fix, but I suspect this to not be sufficient : in some cases the delegate will just have been replaced and here either we will re-execute the old version, either it will throw an `HotReloadException` (that we need to catch to restore the previous state to not break all the nav engine).

src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
```diff
@@ -0,0 +1,121 @@
...
+
+/// <summary>
+/// C# hot-reload handler that re-invokes the <c>viewRouteBuilder</c> delegate
+/// registered via <see cref="ServiceCollectionExtensions.AddNavigation"/> and
+/// rebuilds <see cref="RouteResolver"/> mappings so that newly added routes
+/// become navigable without restarting the app.
+/// </summary>
+internal static class NavigationRouteUpdateHandler
+{
+	private static readonly List<NavigationRouteContext> _contexts = new();
```
dr1rrb: You should use an `ImmutableList` with `ImmutableInterlocked.Update`. This would remove the need to lock everywhere and need to clone (`.ToArray()`)n on each enumeration. We are not expecting lot of updates of that list, it will be more performant.

dr1rrb (approved):

PR #3094, opened 2026-05-01
fix: Re-cascade parent route on child region init instead of ControlInitialize

vatsashah45:

## PR Type

- Bugfix
- Feature

## What is the new behavior?

This pull request addresses a subtle navigation issue in the region-based navigation system, specifically to prevent incorrect route re-cascading after navigation events versus Hot Reload (HR) reloads. The main change introduces logic to distinguish between these scenarios, ensuring that child regions do not incorrectly override their restored routes after navigation. The changes add a suppression flag and supporting methods to propagate this intent through the region tree.

Key changes:

**Navigation Re-cascade Suppression Logic:**

* Added a `_suppressReCascadeOnReload` flag to `NavigationRegion` to indicate when a region is being unloaded due to navigation rather than Hot Reload. This flag prevents the region from re-cascading the parent route upon reload, which would otherwise overwrite the correct child route set by navigation logic.
* Implemented the static method `SuppressReCascadeOnDescendants` in `NavigationRegion`, which recursively marks all descendant regions with the suppression flag. This method is called from `FrameNavigator` before clearing child regions during navigation.

**Region Lifecycle Handling:**

* Updated `ViewUnloaded` in `NavigationRegion` to set a `_wasUnloaded` flag, making it possible to distinguish between first-time loads and reloads after unloads.
* Enhanced the `HandleLoaded` method in `NavigationRegion` to check the new flags and only trigger a route cascade from the parent if the region was reloaded due to Hot Reload, not navigation. After handling, the suppression flag is reset.

**Minor Code Cleanups:**

* Minor whitespace and formatting adjustments for clarity.

These changes ensure that navigation state is restored correctly after navigation events, avoiding accidental overrides during region reloads, and improve the reliability of complex navigation scenarios.

## PR Checklist

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/Navigators/ControlNavigator.cs
+++ b/src/Uno.Extensions.Navigation.UI/Navigators/ControlNavigator.cs
@@ -141,6 +141,11 @@ public abstract class ControlNavigator : Navigator
 {
 	public virtual bool CanGoBack => false;
 
+	// Tracks whether this navigator has received navigation via the normal
+	// route cascade. Used to detect XAML HR re-creation where the cascade
+	// doesn't re-fire for newly created child navigators.
+	private bool _initialNavigationReceived;
+
 	protected ControlNavigator(
 		ILogger logger,
 		IDispatcher dispatcher,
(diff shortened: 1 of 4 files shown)


src/Uno.Extensions.Navigation.UI/Navigators/PanelVisiblityNavigator.cs
```diff
@@ -6,6 +6,10 @@ public class PanelVisiblityNavigator : ControlNavigator<Panel>
 {
 	public const string NavigatorName = "Visibility";
 
+	// Tracks whether Show() has been called by the normal route cascade.
+	// Used to detect when initial content was missed during XAML HR.
+	private bool _showCalled;
```
dr1rrb: From what I understand the `Show()` is to be invoked by the "normal route cascade" ... so I would assume that the fix for HR support would be in that flow, not a patch in the navigator?

src/Uno.Extensions.Navigation.UI/Navigators/ControlNavigator.cs
```diff
@@ -198,10 +203,62 @@ protected virtual NavigationRequest AdjustRequestForChildNavigation(NavigationRe
 
 	public virtual void ControlInitialize()
 	{
+		_initialNavigationReceived = false;
+		_ = DeferredHrRouteCascadeAsync();
```
dr1rrb: Still the same issue, I don't  get why we need to add this in a `ControlInitialize`. We are replicating a navigation request that normally occurs on init ... we need to make that original code path to be re-executed on init.

dr1rrb (approved):

dr1rrb (approved):

PR #3090, opened 2026-04-30
chore: add first mvux HR tests

rajamatt:
## PR Type

What kind of change does this PR introduce?

- Feature

## What is the current behavior?

The reactive hot-reload test class (`Given_HotReload` in `Uno.Extensions.Reactive.UI.Tests`) only covers a bare static method body update (`HotReloadTarget.GetValue`). There is no HR test coverage for the MVUX scenarios described in #2906.

## What is the new behavior?

Adds seven MVUX hot-reload runtime tests to the existing `Given_HotReload` class:

- **`When_UpdateMvuxFeedSource_Then_NewViewModelReflectsUpdate`** — verifies both the original and a new VM pick up an HR'd method body change.
- **`When_RemoveAndReAddFeedProperty_Then_BindingsWork`** — removes then re-adds an `IFeed<string>` property via HR; verifies bindings restore. (#2906)
- **`When_RemoveAndReAddListFeedProperty_Then_BindingsWork`** — same for `IListFeed<string>` bound to a ListView. (#2906)
- **`When_RemoveAndReAddStateProperty_Then_BindingsWork`** — verifies State value update, then remove/re-add cycle for `IState<string>`. (#2906)
- **`When_RemoveAndReAddMultipleProperties_Then_AllBindingsWork`** — removes and re-adds multiple Feed/ListFeed properties in one cycle. (#2906)
- **`When_ConvertFeedToState_Then_BindingsWork`** — converts an `IFeed<string>` to `IState<string>` via HR; verifies bindings survive the type change.
- **`When_ChangeFeedSyntax_Then_BindingsWork`** — cycles Feed.Async → State.Async → Feed.Async across HR deltas; verifies each transition.

Uses dedicated model files per scenario to avoid delta cache interference between tests. Includes `HotReload.Spec.md` documenting all scenarios and harness constraints.

## Files changed

**New:**
- `src/Uno.Extensions.Reactive.UI.Tests/MvuxHotReloadTarget.cs` — HR edit target returning `"original"`
- `src/Uno.Extensions.Reactive.UI.Tests/MvuxHotReloadModel.cs` — MVUX model with `IFeed<string>` reading from target
- `src/Uno.Extensions.Reactive.UI.Tests/MvuxHotReloadFeedRemoveModel.cs` — MVUX model with `IFeed<string>` for remove/re-add
[...]

PR #3095, opened 2026-05-04
fix: HR support for AsyncFeed to refresh on source assembly updates

rajamatt:
## PR Type

- Bug fix

`AsyncFeed<T>` (used by `Feed.Async()`, `State.Async()`, and `ListState.Async()`) has no hot-reload support. When a user edits a model property body, a helper class the property body calls into, or the property's declaring model itself, the feed does not re-evaluate — the UI stays stale. Additionally, ViewModels constructed *after* a hot-reload delta has fired bind to model instances whose lambdas still resolve to pre-update IL.

Three underlying issues:

1. **No subscription to metadata updates.** `AsyncFeed<T>` never observes `HotReloadService.ApplicationUpdated`, so an HR delta to a property's source assembly is invisible to the running feed.

2. **Cross-type indirection is not tracked.** Even with a per-type subscription, `Feed.Async(async ct => Helper.Get())` would not refresh when `Helper.cs` is edited — `Helper` is not the lambda's declaring type. There is no record of *which* types the lambda body actually calls into.

3. **Post-HR `new ViewModel()` is stale.** `[CreateNewOnMetadataUpdate]` causes EnC to produce a fresh shadow type generation per HR cycle, but a `new TBindable()` issued after a delta still constructs `new TModel()` of the *original* type — whose Feed/State lambdas resolve to pre-update IL.

## What is the new behavior?

- **Type-level HR refresh.** `AsyncFeed<T>` subscribes to `HotReloadService.ApplicationUpdated` and re-executes the data provider when the lambda's *declaring type* (not the entire assembly) is in the updated set. A new `HotReloadSupport.AsyncFeed` flag gates this.

- **Cross-type dependency tracking.** A new `FeedDependencyRegistry` maps `(model type, member name) → Type[]` of types the property body references. The MVUX source generator walks each feed/state property body via the semantic model, collects same-assembly type references, and emits a `[ModuleInitializer]` per model that registers those dependencies at assembly load. `AsyncFeed<T>` unions the registered dependencies with the lambda's declaring type to decide whether an HR delta should fire a refresh.

- **Shadow-redirect for post-HR ViewModels.** Generated bindable ctors now invoke `__Reactive_TrySelfHotPatch` at end-of-construction. If `HotReloadService` has recorded a shadow generation for the model since startup, the bindable is rerouted onto the latest generation via the same `HotPatch` path that an HR delta would use on a pre-existing instance — so a fresh ViewModel observes post-HR lambdas.

- **Analyzer reference in `Uno.Extensions.RuntimeTests`.** The MVUX generator is now referenced as an analyzer so the dependency-registry module initializers are emitted for test-project models.

## Files changed

**New:**
- `src/Uno.Extensions.Reactive/Core/HotReload/FeedDependencyRegistry.cs` — `(Type, string member) → Type[]` registry populated by source-generated module initializers; queried by `AsyncFeed<T>`

**Modified — runtime (Reactive):**
- `src/Uno.Extensions.Reactive/Sources/AsyncFeed.cs` — `_sourceType` and `_userMethod` fields, ctor params; `BuildDependentTypeSet` (declaring type ∪ registry entries); `TryGetUserMemberName` parsing of compiler-gen lambda names; `GetOriginalRootType` for EnC shadow resolution; `ApplicationUpdated` subscription with race-safe unsubscribe
- `src/Uno.Extensions.Reactive/Config/HotReloadSupport.cs` — new `AsyncFeed = 1 << 3` flag
- `src/Uno.Extensions.Reactive/Core/Feed.cs`, `Feed.T.cs`, `State.cs`, `State.T.cs`, `ListState.T.cs` — pass `sourceType: vp.Method.DeclaringType` and `userMethod: vp.Method` from the user's original delegate (before any `SomeOrNone…` wrapping) into the AsyncFeed constructors
- `src/Uno.Extensions.Reactive/Core/HotReload/HotReloadService.cs` — `_latestShadow` dictionary tracking the latest shadow generation per original type; `GetLatestShadowType` accessor; `GetOriginalType` widened to `internal` so `AsyncFeed` and `BindableViewModelBase` can call it
- `src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.HotReload.cs` — new `__Reactive_TrySelfHotPatch(Type originalModelType)` method that, if a shadow generation has been registered, runs `HotPatch` to redirect the bindable's model to the freshest type

**Modified — generator:**
- `src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs` — `GenerateFeedDependencies` emits `[ModuleInitializer]` per model invoking `FeedDependencyRegistry.Register(...)`; `ExtractTypeDependencies` walks property body via semantic model collecting same-assembly type references; bindable ctor emits `base.__Reactive_TrySelfHotPatch(typeof(TModel))`
- `src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs` — same `__Reactive_TrySelfHotPatch` emission for the legacy bindable generator

**Modified — test infrastructure:**
- `src/Uno.Extensions.RuntimeTests/Uno.Extensions.RuntimeTests/Uno.Extensions.RuntimeTests.csproj` — adds `Uno.Extensions.Reactive.Generator` as `OutputItemType="Analyzer"` so the dependency-registry initializers are generated for test models
- `src/Uno.Extensions.Reactive.UI.Tests/Given_HotReload.cs` and 10 new `Mvux*Model.cs` test fixtures — runtime test harness covering all hot-reload scenarios (cross-type feed source update, property remove/re-add for Feed/ListFeed/State, multi-property remove/re-add, Feed↔State conversion, AsyncEnumerable round-trips, state value update with new VM)

## Other information

Validated locally with the MVUX hot-reload runtime test suite — **all 11 tests pass** (`Given_HotReload`):

- `When_UpdateMethodBody_Then_MetadataUpdateReceived`
- `When_UpdateMvuxFeedSource_Then_NewViewModelReflectsUpdate` (cross-type)
- `When_UpdateStateValue_Then_NewViewModelReflectsUpdate`
- `When_RemoveAndReAdd{Feed,ListFeed,State}Property_Then_BindingsWork`
- `When_RemoveAndReAddMultipleProperties_Then_AllBindingsWork`
- `When_ConvertFeedToState_Then_BindingsWork`
- `When_AsyncEnumerableFeed_RemoveAndReAdd_Then_BindingsWork`
- `When_ChangeFeedAsyncToAsyncEnumerable_Then_BindingsWork`
- `When_ChangeFeedSyntax_Then_BindingsWork`

The filter is type-level (per review feedback) rather than assembly-wide, so unrelated edits in the same assembly do not trigger spurious refreshes. The cross-type registry lookup gracefully degrades to "declaring type only" when the user method is not a compiler-generated lambda (e.g., a method group reference).

### DIFF
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/MappedMembers/CommandFromMethod.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/MappedMembers/CommandFromMethod.cs
@@ -217,7 +217,7 @@ internal partial record CommandFromMethod : IMappedMember
 			return result.ToString();
 		};
 
-		return @$"{Name} ??= new {NS.Commands}.AsyncCommand(
+		return @$"{Name} = new {NS.Commands}.AsyncCommand(
 					nameof({Name}),
 					new {NS.Commands}.CommandConfig[]
 					{{
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/MappedMembers/CommandFromMethod.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/MappedMembers/CommandFromMethod.cs
@@ -217,7 +217,7 @@ internal partial record CommandFromMethod : IMappedMember
 			return result.ToString();
 		};
 
-		return @$"{Name} = new {NS.Commands}.AsyncCommand(
+		return @$"{Name} ??= new {NS.Commands}.AsyncCommand(
 					nameof({Name}),
 					new {NS.Commands}.CommandConfig[]
 					{{
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs
@@ -139,7 +139,7 @@ internal class ViewModelGenTool_2 : ICodeGenTool
 						{(hasBaseType ? "" : $@"// If a hot-reload delta was applied BEFORE this instance was constructed,
 							// `new {model.Name}()` produced an instance of the original type whose lambdas
 							// still resolve to the pre-update IL. Redirect to the latest shadow generation.
-							base.__Reactive_TrySelfHotPatch(typeof({model.ToFullString()}));").Align(6)}
+							{NS.Core}.HotReload.HotReloadService.TryHotPatch(this);").Align(6)}
 					}}
 
 					#region Hot-reload support
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs
@@ -113,6 +113,8 @@ internal class ViewModelGenTool_2 : ICodeGenTool
 								if ({NS.Config}.FeedConfiguration.EffectiveHotReload.HasFlag({NS.Config}.HotReloadSupport.State))
 								{{
 									__reactiveModelArgs = new (Type, string, object?)[] {{ {ctor.Parameters.Select(p => $"(typeof({p.Type.ToFullString()}), \"{p.Name}\", {p.Name} as object)").JoinBy(", ")} }};
+									// Self-patch from the outermost public ctor — after args are captured. Earlier (base/protected ctor) breaks ctor-deps + inheritance.
+									{NS.Core}.HotReload.HotReloadService.TryHotPatch(this);
 								}}
 							}}")
 						.Align(5)}
@@ -135,11 +137,6 @@ internal class ViewModelGenTool_2 : ICodeGenTool
 							{{
 								npc.PropertyChanged += __Reactive_OnModelPropertyChanged;
 							}}" : "").Align(6)}
-
-						{(hasBaseType ? "" : $@"// If a hot-reload delta was applied BEFORE this instance was constructed,
-							// `new {model.Name}()` produced an instance of the original type whose lambdas
-							// still resolve to the pre-update IL. Redirect to the latest shadow generation.
-							{NS.Core}.HotReload.HotReloadService.TryHotPatch(this);").Align(6)}
 					}}
 
 					#region Hot-reload support
@@ -177,7 +174,8 @@ internal class ViewModelGenTool_2 : ICodeGenTool
 
 						dynamic {N.Ctor.Model} = updatedModel;
 
-						{N.Ctor.Model}.__reactiveBindableViewModel = this;
+						// `(dynamic)this` — required when inherited models shadow this field with a more-derived type.
+						{N.Ctor.Model}.__reactiveBindableViewModel = (dynamic)this;
 
 						{members.Select(member => $@"try
 							{{
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs
@@ -135,6 +135,11 @@ internal class ViewModelGenTool_2 : ICodeGenTool
 							{{
 								npc.PropertyChanged += __Reactive_OnModelPropertyChanged;
 							}}" : "").Align(6)}
+
+						{(hasBaseType ? "" : $@"// If a hot-reload delta was applied BEFORE this instance was constructed,
+							// `new {model.Name}()` produced an instance of the original type whose lambdas
+							// still resolve to the pre-update IL. Redirect to the latest shadow generation.
+							base.__Reactive_TrySelfHotPatch(typeof({model.ToFullString()}));").Align(6)}
 					}}
 
 					#region Hot-reload support
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
@@ -163,7 +163,7 @@ internal class ViewModelGenTool_3 : ICodeGenTool
 						// still resolve to the pre-update IL. Redirect to the latest shadow generation
 						// (re-using the same per-state hot-swap path that an HR delta would trigger
 						// on a pre-existing instance) so this fresh bindable observes the updated values.
-						base.__Reactive_TrySelfHotPatch(typeof({model.ToFullString()}));
+						{NS.Core}.HotReload.HotReloadService.TryHotPatch(this);
 						#endif
 					}}
 
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
@@ -148,6 +148,15 @@ internal class ViewModelGenTool_3 : ICodeGenTool
 							npc.PropertyChanged += __Reactive_OnModelPropertyChanged;
 						}}
 						#endif
+
+						#if {!hasBaseType} // !hasBaseType
+						// If a hot-reload delta was applied BEFORE this instance was constructed,
+						// `new {model.Name}()` produced an instance of the original type whose lambdas
+						// still resolve to the pre-update IL. Redirect to the latest shadow generation
+						// (re-using the same per-state hot-swap path that an HR delta would trigger
+						// on a pre-existing instance) so this fresh bindable observes the updated values.
+						base.__Reactive_TrySelfHotPatch(typeof({model.ToFullString()}));
+						#endif
 					}}
 
 					#region Hot-reload support
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
@@ -131,6 +131,8 @@ internal class ViewModelGenTool_3 : ICodeGenTool
 								if ({NS.Config}.FeedConfiguration.EffectiveHotReload.HasFlag({NS.Config}.HotReloadSupport.State))
 								{{
 									__reactiveModelArgs = new (Type, string, object?)[] {{ {ctor.Parameters.Select(p => $"(typeof({p.Type.ToFullString()}), \"{p.Name}\", {p.Name} as object)").JoinBy(", ")} }};
+									// Self-patch from the outermost public ctor — after args are captured. Earlier (base/protected ctor) breaks ctor-deps + inheritance.
+									{NS.Core}.HotReload.HotReloadService.TryHotPatch(this);
 								}}
 							}}")
 						.Align(5)}
@@ -156,15 +158,6 @@ internal class ViewModelGenTool_3 : ICodeGenTool
 							npc.PropertyChanged += __Reactive_OnModelPropertyChanged;
 						}}
 						#endif
-
-						#if {!hasBaseType} // !hasBaseType
-						// If a hot-reload delta was applied BEFORE this instance was constructed,
-						// `new {model.Name}()` produced an instance of the original type whose lambdas
-						// still resolve to the pre-update IL. Redirect to the latest shadow generation
-						// (re-using the same per-state hot-swap path that an HR delta would trigger
-						// on a pre-existing instance) so this fresh bindable observes the updated values.
-						{NS.Core}.HotReload.HotReloadService.TryHotPatch(this);
-						#endif
 					}}
 
 					#region Hot-reload support
@@ -189,7 +182,9 @@ internal class ViewModelGenTool_3 : ICodeGenTool
 							}}
 						}}
 
-						((dynamic)updatedModel).__reactiveBindableViewModel = this;
+						// `(dynamic)this` — each model partial shadows `__reactiveBindableViewModel` with its own
+						// bindable type; DLR resolves LHS on runtime type and needs RHS runtime type too.
+						((dynamic)updatedModel).__reactiveBindableViewModel = (dynamic)this;
 						base.RegisterDisposable((global::System.IAsyncDisposable)updatedModel);
 
 						__Reactive_BindableInitializeForUpdatedModel(updatedModel, {NS.Core}.SourceContext.GetOrCreate(updatedModel));
@@ -240,7 +235,8 @@ internal class ViewModelGenTool_3 : ICodeGenTool
 					private void __Reactive_OnModelPropertyChanged(object? sender, global::System.ComponentModel.PropertyChangedEventArgs args)
 						=> base.RaisePropertyChanged(args.PropertyName);
 
-					public {(hasBaseType ? "new ":"")}{model.ToFullString()} {N.Model} => ({model.ToFullString()}) __reactiveModel;
+					// `Unsafe.As<T>` — after HR, `__reactiveModel` holds an EnC shadow not assignable to the original. Layout is shared, dispatch is virtual.
+					public {(hasBaseType ? "new ":"")}{model.ToFullString()} {N.Model} => global::System.Runtime.CompilerServices.Unsafe.As<{model.ToFullString()}>(__reactiveModel!);
 
 					{members.Select(member => member.GetDeclaration()).Align(5)}
 				}}");
--- a/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
+++ b/src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
@@ -2,8 +2,11 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Uno.Extensions.Generators;
 using Uno.Extensions.Reactive.Config;
 using static Microsoft.CodeAnalysis.Accessibility;
@@ -41,6 +44,11 @@ internal class ViewModelGenTool_3 : ICodeGenTool
 		{
 			yield return ($"{model}.ViewModel", GenerateViewModel(model));
 			yield return ($"{model}.Reactive", GeneratePartialModel(model));
+
+			if (GenerateFeedDependencies(model) is { } feedDependenciesCode)
+			{
+				yield return ($"{model}.FeedDependencies", feedDependenciesCode);
+			}
 		}
 
 		foreach (var (type, code) in _bindables.Generate())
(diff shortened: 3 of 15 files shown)


src/Uno.Extensions.Reactive/Sources/AsyncFeed.cs
```diff
@@ -48,6 +54,21 @@ public IAsyncEnumerable<Message<T>> GetSource(SourceContext context, Cancellatio
...
+			{
+				if (types.Any(t => t.Assembly == _sourceAssembly))
+				{
+					var refreshedVersion = RefreshToken.InterlockedIncrement(ref current);
+					loadRequests.SetNext(refreshedVersion);
+				}
+			}
+
+			HotReloadService.ApplicationUpdated += OnApplicationUpdated;
+			ct.Register(() => HotReloadService.ApplicationUpdated -= OnApplicationUpdated);
```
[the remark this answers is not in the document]
rajamatt: Tests are on another branch, this is a fix for one of the failing tests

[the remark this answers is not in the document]
rajamatt: I think the handler is cleaned up anyways

src/Uno.Extensions.Reactive/Config/FeedConfiguration.cs
```diff
@@ -39,7 +39,12 @@ internal static HotReloadSupport? HotReload
 	/// Gets the current configuration of the hot-reload support for the MVUX framework.
 	/// </summary>
 	[EditorBrowsable(EditorBrowsableState.Advanced)]
-	public static HotReloadSupport EffectiveHotReload => _effectiveHotReload ??= HotReload ?? HotReloadSupport.Disabled;
+	public static HotReloadSupport EffectiveHotReload => _effectiveHotReload ??= HotReload ?? DetectHotReloadSupport();
+
+	private static HotReloadSupport DetectHotReloadSupport()
+		=> string.Equals(Environment.GetEnvironmentVariable("DOTNET_MODIFIABLE_ASSEMBLIES"), "debug", StringComparison.OrdinalIgnoreCase)
+			? HotReloadSupport.Enabled
+			: HotReloadSupport.Disabled;
```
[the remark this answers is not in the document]
rajamatt: Yes not sure about this.. @dr1rrb let me know what you think. It seems like a hacky workaround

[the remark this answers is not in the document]
rajamatt: Ah, it just needed a ProjectReference to the mvux generator

src/Uno.Extensions.Reactive/Sources/AsyncFeed.cs
```diff
@@ -48,6 +54,21 @@ public IAsyncEnumerable<Message<T>> GetSource(SourceContext context, Cancellatio
...
 		localRefreshTask?.ContinueWith(TryComplete, TaskContinuationOptions.ExecuteSynchronously);
 
+		if (FeedConfiguration.EffectiveHotReload.HasFlag(HotReloadSupport.DynamicFeed) && _sourceAssembly is not null)
+		{
+			void OnApplicationUpdated(Type[] types)
+			{
+				if (types.Any(t => t.Assembly == _sourceAssembly))
+				{
+					var refreshedVersion = RefreshToken.InterlockedIncrement(ref current);
+					loadRequests.SetNext(refreshedVersion);
```
dr1rrb: Basically this means that when we HR an app, we will refresh all feeds of that app ... so far we considered that we should refresh only the feeds that are consumed in a `State` (which is always the case for bindings). You mentioned it's solving an issue with `When_UpdateMvuxFeedSource_Then_NewViewModelReflectsUpdate`, but this should be covered by the `State` refresh on HR, can you provide more details on which step it's  failing ?

dr1rrb: For ref https://github.com/unoplatform/uno.extensions/blob/main/src/Uno.Extensions.Reactive/Core/Internal/StateImpl.cs#L77

dr1rrb: Ok got it ... it's because we are not updating the model itself, so the state is not "hot-swapped" and we need to refresh it. We should then use the type as filter, not the whole assembly! Also we need to create a new flag `HotReloadSupport.AsyncFeed` for that.

Note: we will need to do the same in all "source feeds"

src/Uno.Extensions.Reactive/Sources/AsyncFeed.cs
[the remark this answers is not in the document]
dr1rrb: This is pertinent, we need to address it!

src/Uno.Extensions.Reactive/Sources/AsyncFeed.cs
rajamatt: Done. 

- Type filter + `HotReloadSupport.AsyncFeed` flag.

src/Uno.Extensions.RuntimeTests/Uno.Extensions.RuntimeTests/Uno.Extensions.RuntimeTests.csproj
```diff
@@ -47,5 +47,6 @@
     <ProjectReference Include="..\..\Uno.Extensions.Reactive.UI.Tests\Uno.Extensions.Reactive.WinUI.Tests.csproj" />
     <ProjectReference Include="..\..\Uno.Extensions.Navigation.UI.Tests\Uno.Extensions.Navigation.WinUI.Tests.csproj" />
     <ProjectReference Include="..\Uno.Extensions.RuntimeTests.Core\Uno.Extensions.RuntimeTests.Core.csproj" />
+    <ProjectReference Include="..\..\Uno.Extensions.Reactive.Generator\Uno.Extensions.Reactive.Generator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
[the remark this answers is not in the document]
rajamatt: Other branch

src/Uno.Extensions.Reactive/Core/HotReload/FeedDependencyRegistry.cs
```diff
@@ -0,0 +1,40 @@
...
+			return null;
+		}
+
+		while (owner.DeclaringType is { } d)
+		{
+			owner = d;
+		}
+
+		return _entries.TryGetValue((owner, memberName), out var deps) ? deps : null;
+	}
```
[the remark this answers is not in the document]
rajamatt: Now maps shadow types back to the original before the lookup

src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_2.cs
```diff
@@ -135,6 +135,11 @@ private string GenerateViewModel(INamedTypeSymbol model)
 							{{
```
rajamatt, line 137: [Placeholder comment]

Fixes for tests from #3090: 
- Fix for test `When_UpdateMvuxFeedSource_Then_NewViewModelReflectsUpdate`:
  - Source-gen `FeedDependencyRegistry` for cross-type lambdas, so for example the following code would be a valid scenario if `Helper.Get()` is modified:

```csharp
Feed.Async(ct => Helper.Get())
```

- Fix for test `When_UpdateStateValue_Then_NewViewModelReflectsUpdate`: 
  - Generated bindable ctors call `__Reactive_TrySelfHotPatch`. Without this, a new ViewModel() created after an HR delta would still bind to the original-type model whose lambdas still execute the pre-update method bodies. It redirects the bindable onto the latest shadow generation so the new instance sees post-HR values.

cc @dr1rrb

src/Uno.Extensions.Reactive/Core/Feed.cs
```diff
@@ -69,8 +69,8 @@ public static IFeed<T> Async<T>(AsyncFunc<T> valueProvider, Signal? refresh = nu
 	public static IFeed<T> Async<T>(AsyncFunc<T?> valueProvider, Signal? refresh = null)
 		where T : struct
 		=> refresh is null
-			? AttachedProperty.GetOrCreate(valueProvider, static vp => new AsyncFeed<T>(vp.SomeOrNone()))
-			: AttachedProperty.GetOrCreate(refresh, valueProvider, static (r, vp) => new AsyncFeed<T>(vp.SomeOrNone(), r));
+			? AttachedProperty.GetOrCreate(valueProvider, static vp => new AsyncFeed<T>(vp.SomeOrNone(), sourceType: vp.Method.DeclaringType, userMethod: vp.Method))
+			: AttachedProperty.GetOrCreate(refresh, valueProvider, static (r, vp) => new AsyncFeed<T>(vp.SomeOrNone(), r, vp.Method.DeclaringType, vp.Method));
```
dr1rrb: We are passing the `<delegate>.Method.DeclaringType` and the `<delegate>.Method` everywhere ... just capture the `<delegate>`. This will also allow us to ensure to **not** access to the `Method` property on the `Delegate` which might involves reflection.

src/Uno.Extensions.Reactive/Sources/AsyncFeed.cs
```diff
@@ -13,15 +17,24 @@ internal sealed class AsyncFeed<T> : IFeed<T>
...
 	}
 
-	public AsyncFeed(AsyncFunc<Option<T>> dataProvider, ISignal? refresh = null)
+	public AsyncFeed(AsyncFunc<Option<T>> dataProvider, ISignal? refresh = null, Type? sourceType = null, MethodInfo? userMethod = null)
 	{
+		_sourceType = sourceType ?? dataProvider.Method.DeclaringType;
+		// `dataProvider` is often a wrapper (`SomeOrNone()`, `async ct => await user(ct)`) — its Method
+		// would point at the wrapper, not the user's lambda. Callers that wrap should pass the user's
+		// original Method so the per-property dependency-registry lookup uses the correct member name.
+		_userMethod = userMethod ?? dataProvider.Method;
```
dr1rrb: As per my comment on the factories, we should only have a single field `Delegate _hotReloadDelegate` with 

```suggestion
	public AsyncFeed(AsyncFunc<Option<T>> dataProvider, ISignal? refresh = null, Delegate rawDataProvider = null)
{
		if (FeedConfiguration.EffectiveHotReload.HasFlag(HotReloadSupport.AsyncFeed)) // Avoid potential leak if HR is disbaled
		{
			_hotReloadDelegate = rawDataProvider  ?? dataProvider;
		}
```

src/Uno.Extensions.Reactive/Sources/AsyncFeed.cs
```diff
@@ -48,6 +62,35 @@ public IAsyncEnumerable<Message<T>> GetSource(SourceContext context, Cancellatio
 
 		localRefreshTask?.ContinueWith(TryComplete, TaskContinuationOptions.ExecuteSynchronously);
 
+		if (FeedConfiguration.EffectiveHotReload.HasFlag(HotReloadSupport.AsyncFeed) && _sourceType is not null)
+		{
+#pragma warning disable IL2026 // Underlying API uses reflection on per-assembly metadata that cannot be statically known.
+			// Build the filter set: the lambda's declaring type, plus any extra types the source
+			// generator recorded as dependencies of this property's body (cross-type calls like
+			// `Feed.Async(async ct => Helper.Get())` — editing Helper.cs needs to refresh too).
+			var dependentTypes = BuildDependentTypeSet();
```
dr1rrb:
```suggestion
		if (_hotReloadDelegate is { Method: { DeclaringType: { } sourceType } sourceMethod })
		{
#pragma warning disable IL2026 // Underlying API uses reflection on per-assembly metadata that cannot be statically known.
			// Build the filter set: the lambda's declaring type, plus any extra types the source
			// generator recorded as dependencies of this property's body (cross-type calls like
			// `Feed.Async(async ct => Helper.Get())` — editing Helper.cs needs to refresh too).
			var dependentTypes = BuildDependentTypeSet(sourceType, sourceMethod);
```

src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.HotReload.cs
```diff
@@ -114,6 +115,47 @@ private void HotPatch(
...
+			return;
+		}
+
+		// Note: even when `shadowType == originalModelType` we still patch — `new TModel()` invoked
+		// from the generated bindable ctor produces an instance of the ORIGINAL type (its IL has
+		// the original ctor token), whose lambdas resolve to pre-HR delegates. Only re-creating the
+		// model via the shadow type's metadata gives us the post-HR lambdas.
+		HotPatch(canonicalOriginal, shadowType);
+	}
+
```
dr1rrb: Why reverting the dependency between the current flow and this new flow? Move all of this in the `HotReloadService`, revert visibility changes in the service (we can also remove the `GetLatestShadowType`), and call `HotReloadService.TryHotPatch(this)` at the end of the ctor.

src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
```diff
@@ -148,6 +156,15 @@ private string GenerateViewModel(INamedTypeSymbol model)
...
 						}}
 						#endif
+
+						#if {!hasBaseType} // !hasBaseType
+						// If a hot-reload delta was applied BEFORE this instance was constructed,
+						// `new {model.Name}()` produced an instance of the original type whose lambdas
+						// still resolve to the pre-update IL. Redirect to the latest shadow generation
+						// (re-using the same per-state hot-swap path that an HR delta would trigger
+						// on a pre-existing instance) so this fresh bindable observes the updated values.
+						base.__Reactive_TrySelfHotPatch(typeof({model.ToFullString()}));
```
dr1rrb: This will not work for sub-classes : the members init that occurs on line 151 will undo the hot-patch, we need to defer it.

src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.HotReload.cs
dr1rrb: Actually, the `HotRelaodService.TryHotPatch` should be invoked at the end of the `InitializeHotReload()` https://github.com/unoplatform/uno.extensions/blob/03e4af63a192c3e855e9f1a34427fb049af0f200/src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.HotReload.cs#L100 which is already invoked in the ctor of the `BindableViewModelBase` https://github.com/unoplatform/uno.extensions/blob/03e4af63a192c3e855e9f1a34427fb049af0f200/src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.cs#L49

src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
dr1rrb: Ok the only problem will be https://github.com/unoplatform/uno.extensions/blob/03e4af63a192c3e855e9f1a34427fb049af0f200/src/Uno.Extensions.Reactive.Generator/Bindables/MappedMembers/CommandFromMethod.cs#L220 we need to replace by

```csharp
return @$"{Name} ??= new {NS.Commands}.AsyncCommand(
```

Then we can do what I suggested here : https://github.com/unoplatform/uno.extensions/blob/03e4af63a192c3e855e9f1a34427fb049af0f200/src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.cs#L49

(The fields and properties will actually be init by the HR.HotPatch, but this is fine as all the `GetInitialization` are now validating that field/member is not initialized before creating it)

rajamatt: Done

src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.HotReload.cs
```diff
@@ -114,6 +115,12 @@ private void HotPatch(
 		__Reactive_UpdateModel(model);
 	}
 
+	internal void HotPatchSelf(
+		Type originalModel,
+		[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]
+		Type updatedModelType)
+		=> HotPatch(originalModel, updatedModelType);
+
```
dr1rrb: We can simply invoke `HotPatch` :)

```suggestion
```

src/Uno.Extensions.Reactive/Core/HotReload/HotReloadService.cs
```diff
@@ -61,10 +76,59 @@ internal static void UpdateApplication(Type[]? types)
...
+			return;
+		}
+
+		// Note: even when `shadowType == modelType` we still patch — `new TModel()` invoked from the
+		// generated bindable ctor produces an instance of the ORIGINAL type (its IL has the original
+		// ctor token), whose lambdas resolve to pre-HR delegates. Only re-creating the model via the
+		// shadow type's metadata gives us the post-HR lambdas.
+		try
+		{
+			target.HotPatchSelf(canonicalOriginal, shadowType);
```
dr1rrb:
```suggestion
			target.HotPatch(canonicalOriginal, shadowType);
```

src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
```diff
@@ -148,6 +156,15 @@ private string GenerateViewModel(INamedTypeSymbol model)
...
 						#endif
+
+						#if {!hasBaseType} // !hasBaseType
+						// If a hot-reload delta was applied BEFORE this instance was constructed,
+						// `new {model.Name}()` produced an instance of the original type whose lambdas
+						// still resolve to the pre-update IL. Redirect to the latest shadow generation
+						// (re-using the same per-state hot-swap path that an HR delta would trigger
+						// on a pre-existing instance) so this fresh bindable observes the updated values.
+						{NS.Core}.HotReload.HotReloadService.TryHotPatch(this);
+						#endif
```
dr1rrb: I still think this is not required, we only need to invoke `HotReloadService.TryHotPatch` at the end of the `InitializeHotReload()`  (and remove this also from the VMGenTool_2). Is there any reason to not do so?

https://github.com/unoplatform/uno.extensions/blob/03e4af63a192c3e855e9f1a34427fb049af0f200/src/Uno.Extensions.Reactive/Presentation/Bindings/BindableViewModelBase.HotReload.cs#L100

cf. https://github.com/unoplatform/uno.extensions/pull/3095#discussion_r3210441717

rajamatt: Okay so I think this is resolved? We found that `TryHotPatch` has to run after `__reactiveModelArgs` is set. So we emit the event at the end of the bindable model ctor. But then there's the problem of inherited models. So we emit the event when all models have been constructed.

src/Uno.Extensions.RuntimeTests/Uno.Extensions.RuntimeTests/Uno.Extensions.RuntimeTests.csproj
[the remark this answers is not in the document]
rajamatt: Tests are in a different PR

src/Uno.Extensions.Reactive/Core/HotReload/FeedDependencyRegistry.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Collections.Concurrent;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Uno.Extensions.Reactive.Core.HotReload;
+
+/// <summary>
+/// Internal registry populated by the Reactive source generator. Maps a (model type, member name)
+/// pair to the set of types whose IL the corresponding feed/state lambda calls into. Used by
+/// <see cref="Sources.AsyncFeed{T}"/> to decide whether a hot-reload metadata update should
```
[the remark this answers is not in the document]
rajamatt: I feel like this is being too precise and not needed

src/Uno.Extensions.Reactive/Config/HotReloadSupport.cs
```diff
@@ -22,6 +22,13 @@ public enum HotReloadSupport
 	/// </summary>
 	State = 1 << 2,
 
+	/// <summary>
+	/// Enables the hot-reload support for async feeds. The feed re-evaluates when the type that declares the
+	/// data provider is updated, OR when any type recorded in the source-generated <see cref="Core.HotReload.FeedDependencyRegistry"/>
```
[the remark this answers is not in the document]
rajamatt: I feel like this is being too precise and not needed

src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
```diff
@@ -223,7 +235,8 @@ protected override sealed void __Reactive_UpdateModel(object updatedModel)
 					private void __Reactive_OnModelPropertyChanged(object? sender, global::System.ComponentModel.PropertyChangedEventArgs args)
 						=> base.RaisePropertyChanged(args.PropertyName);
 
-					public {(hasBaseType ? "new ":"")}{model.ToFullString()} {N.Model} => ({model.ToFullString()}) __reactiveModel;
+					// `Unsafe.As<T>` — after HR, `__reactiveModel` holds an EnC shadow not assignable to the original. Layout is shared, dispatch is virtual.
+					public {(hasBaseType ? "new ":"")}{model.ToFullString()} {N.Model} => global::System.Runtime.CompilerServices.Unsafe.As<{model.ToFullString()}>(__reactiveModel!);
```
rajamatt, line 239: Not sure about use of Unsafe.As... This was the agent's solution to seeing errors in the debug console when testing in a separate uno app 😅 @dr1rrb

src/Uno.Extensions.Reactive.Generator/Bindables/MappedMembers/CommandFromMethod.cs
```diff
@@ -217,7 +217,7 @@ string GetDeconstruct(string args, string ct, string? viewArg = null)
 			return result.ToString();
 		};
 
-		return @$"{Name} = new {NS.Commands}.AsyncCommand(
+		return @$"{Name} ??= new {NS.Commands}.AsyncCommand(
```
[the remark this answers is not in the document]
rajamatt: @dr1rrb Thoughts here?

src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.cs
rajamatt: The error was 
```
fail: Uno.UI.DataBinding.BindingPath+BindingItem[0]
        Failed to get the source value for [Title]
        System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
         ---> System.InvalidCastException: Unable to cast object of type
              'Test6dot6.Presentation.MainModel#1' to type
              'Test6dot6.Presentation.MainModel'.
           at Test6dot6.Presentation.MainViewModel.get_Model()
           at Test6dot6.Presentation.MainViewModel.get_Title()
```

The agent's thoughts:

> **What was happening**: the generator emitted `public MainModel Model => (MainModel) __reactiveModel;`. After HR, `__reactiveModel` held the EnC shadow `MainModel#1`, which the CLR sees as a *distinct* type (the `MetadataUpdateOriginalTypeAttribute` connects them at the metadata layer but doesn't affect runtime cast semantics). So the explicit cast threw on every access — and `{Binding Title}` traversed `Bindable.Title → Model.Title`, so the `Title` binding broke.
> 
> The fix changed it to `Unsafe.As<MainModel>(__reactiveModel!)` which reinterprets the reference without a runtime type check. EnC guarantees shadow types share the field layout of their original, and member access dispatches virtually on the runtime type, so the reinterpret is sound.

dr1rrb: It's fine here

src/Uno.Extensions.Reactive.Generator/Bindables/MappedMembers/CommandFromMethod.cs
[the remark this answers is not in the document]
dr1rrb: Good point, but it means that in debug (or actually when if `({NS.Config}.FeedConfiguration.EffectiveHotReload.HasFlag({NS.Config}.HotReloadSupport.State))`) but without this, it will always create the command twice (one in VM ctor, and one for the `TryHotPatch`) ...I think this will enough for now, just remove `??`

dr1rrb (approved):

PR #3101, opened 2026-05-11
fix: yield to dispatcher in EnsureChildRegionsAreLoaded when children 

vatsashah45:

## PR Type

- Bugfix

In certain hosting scenarios, specifically when an app is loaded in-process via AssemblyLoadContext (as Studio Live does), the initial navigation fails on first load with errors like:
- Navigation to 'Main' failed: Show() returned null. No matching view was found or created.
- Region 'Dashboard' has no children to forward request to

This happens because Navigator.CoreNavigateAsync() calls EnsureChildRegionsAreLoaded(), which waits for the current region's view to load via CheckLoadedAsync(), but child NavigationRegion instances attach to parent.Children during their own Loaded event handler. In ALC hosting, these Loaded events may be queued on the dispatcher but not yet processed when CheckLoadedAsync) returns, so Region.Children.Count is 0 and navigation fails immediately.

On a second Build+Publish cycle the visual tree is already constructed and child regions are attached, so navigation succeeds — making this a first-load-only race condition.

## What is the new behavior?

After CheckLoadedAsync() returns in EnsureChildRegionsAreLoaded(), if Region.Children is still empty despite the view being loaded, the method now yields to the dispatcher (up to 5 times via Task.Yield() to allow pending Loaded events to fire and child regions to complete attachment.

This is zero-cost in normal scenarios — the yield loop only executes when Children.Count == 0 && View.IsLoaded == true, which doesn't occur in standard hosting where children attach synchronously during the view's load cycle.

## PR Checklist

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/Navigator.cs
+++ b/src/Uno.Extensions.Navigation.UI/Navigator.cs
@@ -882,6 +882,27 @@ public class Navigator : INavigator, IInstance<IServiceProvider>
 		// This is required to ensure nested elements (eg Content in a ContentControl)
 		// are loaded. This will ensure the Children collection is correctly populated
 		await CheckLoadedAsync();
+
+		// Child NavigationRegions attach themselves to parent.Children during their
+		// Loaded event handler (HandleLoading → AssignParent). In some hosting
+		// scenarios (e.g., in-process AssemblyLoadContext loading), these Loaded
+		// events may be queued on the dispatcher but not yet processed when this
+		// method returns. Re-schedule on the dispatcher to ensure its queue is
+		// drained and pending attachments complete before the caller checks
+		// Region.Children.
+		if (Region.Children.Count == 0 && Region.View?.IsLoaded == true)
+		{
+			const int maxAttempts = 5;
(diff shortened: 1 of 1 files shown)


src/Uno.Extensions.Navigation.UI/Navigator.cs
```diff
@@ -882,6 +882,26 @@ private async Task EnsureChildRegionsAreLoaded()
...
+		if (Region.Children.Count == 0 && Region.View?.IsLoaded == true)
+		{
+			const int maxYields = 5;
+			for (var attempt = 0; attempt < maxYields && Region.Children.Count == 0; attempt++)
+			{
+				if (Logger.IsEnabled(LogLevel.Trace))
+				{
+					Logger.LogTraceMessage($"Children not yet attached after view loaded (attempt {attempt + 1}/{maxYields}), yielding to dispatcher");
+				}
+				await Task.Yield();
```
dr1rrb: If it's dispatching issue, we should not `Task.Yield` but re-schedule on dispatcher (in `Idle` ?). The Dispatcher might be busy to load something huge, the `Task.Yield` might all completes before the dispatcher process anything in it queue.

dr1rrb (approved):

PR #3109, opened 2026-05-21
fix: Enhance navigation handling with hot-reload support

kazo0:
## Summary

Makes navigation route resolution survive C# hot-reload cycles and recover from initial navigations that fire before their target view type is loaded (e.g. Studio Live's scaffolding pipeline). Adds spec + regression test.

## Problem

After scaffolding a tabbed app via Studio Live's hot-reload-driven generation, the content area above the `TabBar` stayed blank forever — the shell rendered, but the `IsDefault` nested page never appeared. A full app rebuild fixed it; subsequent hot-reload deltas did not.

Two root causes:

1. **`NavigationRouteContext.Resolver` was never assigned.** `ServiceCollectionExtensions.AddNavigation` registered `IRouteResolver` via a factory delegate that side-effected `ctx.Resolver = routeResolver`. `HostBuilderExtensions.UseNavigation` then ran `services.AddSingleton<IRouteResolver, MappedRouteResolver>()` afterwards — last-registration-wins, so the factory never executed. Every `[MetadataUpdateHandler] UpdateApplication` bailed at `RebuildRoutes: resolver is null`.

2. **Initial navigation requests fired before their target type was loaded** had no recovery path. The `Show()` call returned null, the request was dropped, and even after the missing type appeared via a subsequent HR delta, nothing replayed the navigation.

### Wiring fix (`NavigationHostedService.cs`, `NavigationRouteContext`)
`StartAsync` now wires `IRouteResolver` into `NavigationRouteContext.Resolver` explicitly, after both have been resolved from DI. Closes the registration-order gap without changing the public DI surface.

### HR resolver rebuild (`NavigationRouteUpdateHandler.cs`, `RouteResolver.Rebuild`)
On `UpdateApplication`:
- Snapshot the view + route registries, clear them, re-invoke the route builder delegate (whose method body may have been replaced).
- Rebuild the resolver's flat lookup tables via the new `RouteResolver.Rebuild()`.
- On builder throw: restore the snapshots so navigation continues against the pre-HR route table.
- Per-context try/catch ensures one failing context doesn't abort the rest; never propagates back into the .NET HR runtime.

### Cascade after rebuild
After a successful rebuild, schedules a deferred (`IDispatcher.TryEnqueue`) walk of the live region tree:
- For each navigator whose `RouteInfo` exposes a nested `IsDefault` route, dispatch navigation to the **first child region** (not the navigator itself) — going one level deeper avoids tearing down the parent's region tree and racing the freshly-added XAML-HR children.
- `FindActiveDescendantNestedRoute` short-circuits the cascade if any descendant already has an active route matching one of the nested paths — preserves a user's selection (e.g. they switched to RegionTwo) instead of clobbering it back to the `IsDefault`.

A second walk (`RetryPendingFailedRequestsFromRoot`) re-issues any pending failed navigation requests on `ControlNavigator` instances — the resolver was just rebuilt, so previously-missing types may now resolve.

### Pending-request slot (`ControlNavigator`)
When `Show()` returns null, the most recent failing `NavigationRequest` is parked in a private `_pendingFailedRequest` slot (single slot, not a queue — newer failures overwrite older). The HR retry walk picks it up. Log de-flooding via `HasPendingFailedRequestFor(route)`: repeated failures of the same route demote to Debug instead of re-warning on every HR poll. A separate Page-subclass-in-FrameView path short-circuits cleanly with `Route.Empty` + `EnsureLoaded()`.

### External entry point (`ScheduleCascadeForAllContexts`)
`internal` API for callers outside the C# HR pipeline (XAML HR doesn't invoke `MetadataUpdateHandler`) to trigger the same cascade walk against existing contexts.

### Fire-and-forget hardening
The cascade and retry dispatch navigation via `TryEnqueue(... _ = NavigateAsync(...))`. Both sites now route through `SafeFireAndForget`, which awaits inside `try/catch`, swallows `OperationCanceledException` for teardown races, and logs other failures at Warning. `NavigationHostedService.StopAsync` similarly guards `Unregister`.

### Public-surface discipline
The pending-request bookkeeping members on `ControlNavigator` (`RememberPendingFailedRequest`, `ClearPendingFailedRequest`, `HasPendingFailedRequestFor`) are `internal` — they're plumbing for the same-assembly generic subclass and the HR handler, not contract for downstream navigators.

### Log hygiene
The new "no children to forward" Warning in `Navigator.cs` now logs `Route.Base` rather than the full `Route.ToString()`, which interpolates `Route.Query()` built from `Data` (potential token/PII surface per AGENTS.md §7).

## Tests

- **`Given_HotReload.When_HRCascadeAfterUserSelection_Then_ActiveRegionPreserved`** — regression test: navigates the panel onto a non-default nested region, applies an HR source change to trigger `UpdateApplication`, asserts the user-selected region is preserved across the cascade walk. Exercises `FindActiveDescendantNestedRoute`.

## Design rationale

See [`specs/2026-05-20 hot-reload-navigation-route-resolver/spec.md`](specs/2026-05-20%20hot-reload-navigation-route-resolver/spec.md).

## Risks / follow-ups

- `_pendingFailedRequest` is a single slot, not a queue — a second failure overwrites the first. Documented in source comments; consistent with the "most recent intent wins" semantic for HR-recovery scenarios.
- `CascadeNewDefaultsFromRoot` targets `region.Children.FirstOrDefault()`. Works for the Frame → page-region shape; composite layouts with named sibling regions may want a more specific selector — left as a follow-up.
- The `IRouteResolver` factory in `ServiceCollectionExtensions.AddNavigation` still side-effects `ctx.Resolver` for parity. The hosted-service assignment makes that side-effect redundant; safe to remove in a follow-up cleanup.

## Test plan

- [ ] Studio Live's tabbed-app scaffold: `IsDefault` page renders without app restart.
- [ ] HR delta to a route delegate body: subsequent navigation reflects the change.
- [ ] User-selected non-default nested region survives an HR delta unrelated to navigation.
- [ ] Pending failed nav from a missing type recovers on the HR delta that adds the type.
- [ ] No regression in non-HR navigation (default behavior unchanged when no `MetadataUpdateHandler` fires).


src/Uno.Extensions.Navigation.UI/Navigator.cs
```diff
@@ -763,7 +764,30 @@ protected virtual async Task<bool> RegionCanNavigate(Route route, RouteInfo? rou
...
+				// Show() returned null and the request is parked waiting for HR to
+				// deliver the missing type — the retry path will re-issue the
+				// navigation once children attach. In that state, demote the warn
+				// to Debug so the bundle doesn't fill up with the same race-symptom
+				// across every HR delta. Real misconfigurations (no pending retry,
+				// no HR in flight) still surface at Warning.
+				var inPendingRetry = (Region.Navigator() as ControlNavigator)?.HasPendingFailedRequest ?? false;
+				if (inPendingRetry)
+				{
+					if (Logger.IsEnabled(LogLevel.Debug)) Logger.LogDebugMessage($"Region '{Region.Name}' has no children to forward request to (route: '{request.Route.Base}', view loaded: {Region.View?.IsLoaded}) — quiet because a pending HR retry is in flight on this region");
```
jeromelaban: Formatting is different from the rest (no `{}`)

src/Uno.Extensions.Navigation.UI/Navigators/ControlNavigator.cs
```diff
@@ -141,6 +201,73 @@ public abstract class ControlNavigator : Navigator
...
+		var pending = _pendingFailedRequest;
+		if (pending is null)
+		{
+			return Task.CompletedTask;
+		}
+
+		if (Logger.IsEnabled(LogLevel.Information))
+		{
+			Logger.LogInformationMessage($"Retrying pending navigation '{pending.Route.Base}' after hot-reload");
+		}
```
jeromelaban: This does not seem like a info-level type, probably just debug.

src/Uno.Extensions.Navigation.UI/Navigators/ControlNavigator.cs
```diff
@@ -57,11 +57,71 @@ protected override Task<bool> RegionCanNavigate(Route route, RouteInfo? routeMap
...
 
+		// If we were in a pending-retry loop for this same route, the success
+		// here is the moment the missing type finally landed via hot-reload.
+		// Log it at Information so the bundle has a clear "recovered" marker
+		// that pairs with the original Warning, rather than the navigation
+		// silently going green with no log trail.
+		if (HasPendingFailedRequestFor(route) && Logger.IsEnabled(LogLevel.Information))
+		{
+			Logger.LogInformationMessage($"Navigation to '{route.Base}' recovered after hot-reload (pending retry succeeded).");
+		}
```
jeromelaban: Same here, debug only

jeromelaban (approved):

PR #3114, opened 2026-05-30
fix: skip HR cascade when no route-registered type was updated

vatsashah45:
GitHub Issue (If applicable): Related to https://github.com/unoplatform/studio.live/issues/2293

## PR Type

- Bugfix

## What is the new behavior?

When a XAML-only hot-reload cycle fires (e.g. a Text or colour change on a page with x:Uid), NavigationRouteUpdateHandler.UpdateApplication received the updated types but unconditionally scheduled a cascade walk after rebuilding the route table.  That cascade re-mounted the page, causing Uno's x:Uid resolver to fire InitializeComponent and overwrite any property values that Hot Design had just written to XAML literals.

Two guard points added:

1. NavigationRouteUpdateHandler.RebuildRoutes — after a successful rebuild, only call ScheduleCascade when at least one of the updatedTypes is registered in the resolver as a view or view-model. A helper HasRouteRegisteredType() does the lookup via FindByView / FindByViewModel; the IL2072 suppression is justified because the calls are read-only type-identity lookups, not construction.

2. NavigationVisibilityUpdateHandler.RestoreState — replaces the unconditional ScheduleCascadeForAllContexts() call with a new ScheduleCascadeForAllContextsIfRouteRelevant(updatedTypes) that applies the same guard before each context's cascade walk.

A null updatedTypes is treated as 'always cascade' to preserve the existing contract for callers that don't have type information.

## PR Checklist

### DIFF
--- a/src/Uno.Extensions.Navigation.UI.Tests/Given_NavigationRouteUpdateHandler.cs
+++ b/src/Uno.Extensions.Navigation.UI.Tests/Given_NavigationRouteUpdateHandler.cs
@@ -1,0 +1,77 @@
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Uno.Extensions.Navigation;
+using Uno.Extensions.Navigation.UI;
+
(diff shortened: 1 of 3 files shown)


dr1rrb (commented):
👍

Are we able to add tests for that ?

dr1rrb (approved):

PR #3117, opened 2026-06-11
fix(navigation): park and resume navigation dropped during tree re-graft

kazo0:
An app comes up permanently blank when its hosting tree is re-grafted while the initial navigation is in flight (e.g. Hot Design re-hosting an ALC-loaded app's content during bootstrap). Spec 006:

- park child-forwards when the region's view is detached or no child regions are attached; resume from NavigationRegion.HandleLoaded on (re-)attach
- park requests arriving while a region is detached (no services); resume on the region's own reload
- host-aware FrameView waits: EnsureLoadedWhileHostAttached gives up when the hosting view leaves the tree instead of hanging the timeout-less wait
- test: Given_NestedNavReloadRecovery (deterministic detach inside Frame.Loaded, re-attach, assert the initial route resumes


src/Uno.Extensions.Navigation.UI/FrameworkElementExtensions.cs
```diff
@@ -113,6 +113,48 @@ internal static async Task Startup(this IServiceProvider services, Func<Task> af
...
+	/// Waits for <paramref name="element"/> to load, but gives up as soon as
+	/// <paramref name="hostView"/> leaves the tree (or is already detached): a child
+	/// element cannot load while its hosting subtree is detached, and
+	/// <see cref="EnsureLoaded"/> has no timeout — waiting on a detached subtree hangs
+	/// the navigation pipeline forever (spec 006: the hosting tree being re-grafted
+	/// mid-navigation, e.g. Hot Design wrapping the app content during bootstrap).
+	/// Callers fall through on <see langword="false"/>; the request is then parked by
+	/// the child-forwarding stage and resumed when the region re-attaches.
+	/// </summary>
+	internal static async Task<bool> EnsureLoadedWhileHostAttached(this FrameworkElement? element, FrameworkElement? hostView)
```
nickrandolph, line 127: This feels quite messy and that adapting the existing EnsureLoaded method to take a cancellationtoken would have been a cleaner approach. If/When the hostView is unloaded the cancellation token could be cancelled, preventing the EnsureLoaded from hanging indefinitely.

kazo0: Reworked in 1e534fb4d as suggested: `EnsureLoaded` now takes a `CancellationToken` (linked with its existing optional timeout), and `EnsureLoadedWhileHostAttached` is reduced to wiring `hostView.Unloaded` → `Cancel()` before delegating. This drops the `TaskCompletionSource`/`Task.WhenAny` race entirely, so there's no orphaned load task left running when the host detaches, and the `IsLoaded` re-check after subscribing closes the detach race window Copilot flagged earlier. Cancellation surfaces as the same `false` fall-through the park/resume path already handles.

PR #3119, opened 2026-06-15
feat(navigation): allow bounding navigation to a portion of the visual tree

Kunal22shah:
Navigation resolves an element's service provider / parent region by walking **up** the visual tree (`DependencyObjectExtensions.ServiceForControl`), and there is currently no way to tell it to stop at a given element. So an element hosted inside another visual tree (e.g. an embedded or hosted subtree) self-wires into the surrounding navigation and can duplicate the nav host (such as a second TabBar).

This treats `Region.Attached="false"` as a **navigation boundary**: when an element has `Region.Attached` set to an explicit *local* `false` **and** its effective value is also `false`, the upward service-provider walk stops there, so that element's subtree isn't attached to the navigation above it.

Reusing the existing `Region.Attached` property keeps the surface minimal — `Attached="true"` opts a subtree into navigation; `Attached="false"` now bounds it out. Requiring the **effective** value to be false (not just the local value) preserves the responsive master-detail idiom, where a pane is declared `Region.Attached="false"` locally and flipped to `true` by a VisualState/Style setter for the wide layout — that pane stays a real region, not a boundary.

`NavigationRegion` also suppresses the "Unable to find service provider for root navigator" warning when a root region legitimately sits below a boundary (the `IsEnabled` check runs first so the boundary tree-walk only happens when we would actually log; a Trace line records the suppression).

## PR Type

- Feature

`ServiceForControl` walks unconditionally to the visual root. An element hosted inside a larger visual tree cannot be detached from the navigation above it, so a hosted nav shell finds the outer service provider, self-promotes to a root navigator, default-navigates, and duplicates the nav host (a second TabBar).

## What is the new behavior?

An element with an explicit, effective `Region.Attached="false"` is a navigation boundary: the service-provider walk stops there and its subtree resolves no region/SP from above. The responsive `false`→`true` (VisualState) idiom is unaffected. New `Given_NavigationBoundary` runtime tests cover: walk stops at an explicit-false boundary, default-false is not a boundary, and effective-true is not a boundary.

## PR Checklist

- [x] Unit/UI tests for the changes have been added (`Given_NavigationBoundary`)
- [x] Associated with an issue (closes studio.live#2740)

### DIFF
--- a/src/Uno.Extensions.Navigation.UI/DependencyObjectExtensions.cs
+++ b/src/Uno.Extensions.Navigation.UI/DependencyObjectExtensions.cs
@@ -68,7 +68,31 @@ public static class DependencyObjectExtensions
             return default;
         }
 
+        // Stop at a navigation boundary: an element with Region.Attached explicitly set to false
+        // detaches its subtree from the navigation above it, so descendants resolve no region/SP.
+        if (element.IsNavigationBoundary())
+        {
+            if (Region.Logger.IsEnabled(LogLevel.Trace))
+            {
+                Region.Logger.LogTraceMessage($"Navigation boundary reached at {element.GetType().Name} (Region.Attached=false); stopping service-provider walk - the subtree below is isolated from the navigation above it.");
+            }
+
+            return default;
+        }
+
         var parent = VisualTreeHelper.GetParent(element);
         return parent.ServiceForControl(searchParent, retrieveFromElement);
     }
+
+    /// <summary>
+    /// True when Region.Attached marks this element as a navigation boundary: its subtree is not
+    /// attached to the navigation above it. Requires a <em>local</em> <c>false</c> (so the default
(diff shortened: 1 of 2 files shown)


src/Uno.Extensions.Navigation.UI/DependencyObjectExtensions.cs
```diff
@@ -68,7 +83,27 @@ public static class DependencyObjectExtensions
             return default;
         }
 
+        // Stop at a navigation boundary: an element with Region.Attached explicitly set to false
+        // detaches its subtree from the navigation above it, so descendants resolve no region/SP.
+        if (element.IsNavigationBoundary())
+        {
+            stoppedAtBoundary = true;
```
nickrandolph: Would seem to be simpler to just log that navigation boundary reached here than having to bubble up a "stoppedAtBoundary" flag.

kazo0 (approved):

PR #3121, opened 2026-06-29
fix(themeservice): apply theme to own root element, not XamlRoot content

### LINKED ISSUE #3120: ThemeService sets RequestedTheme on XamlRoot's content instead of its own root element (breaks theme switching when hosted in foreign XamlRoot / ALC)
## Current behavior

`Uno.Extensions.Toolkit.ThemeService` (consumed via `UseThemeSwitching` / `IThemeService`, registered as `ScopedThemeService`) applies the theme to `RootElement.XamlRoot.Content` rather than to `RootElement` itself. When an app's content is hosted under a `XamlRoot` it doesn't own — e.g. a secondary app loaded into a collectible `AssemblyLoadContext` whose `Window.Content` is re-parented into a host's single shared `XamlRoot` — the theme is applied to the **host's** root visual instead of the app's. Toggling the theme re-themes the host while the hosted app does not change.

## Expected behavior

When theme switching is toggled via `IThemeService.SetThemeAsync(...)`, the hosted app's theme should change, and `ThemeChanged` should fire correctly. The behavior should be consistent whether the app is standalone or hosted under a foreign `XamlRoot`.

1. App registers theme switching (`UseThemeSwitching()`), with a light/dark toggle calling `IThemeService.SetThemeAsync(...)`.
2. Run the app **hosted inside another app's XamlRoot** — a secondary-app/ALC scenario where the inner app's `Window.Content` is mounted into a host's `XamlRoot`.
3. Toggle the theme.

**Expected:** the hosted app re-themes.
**Actual:** the host re-themes; the hosted app does not. `ThemeChanged` also fails to fire.

Note: Standalone apps are unaffected because there `window.Content == XamlRoot.Content`, so the bug is invisible.

vatsashah45:
GitHub Issue: closes #3120

## PR Type

- Bugfix

`Uno.Extensions.Toolkit.ThemeService` (consumed via `UseThemeSwitching` / `IThemeService`, registered as `ScopedThemeService`) applies the theme to `RootElement.XamlRoot.Content` — the **root visual of the XamlRoot** — rather than to `RootElement` itself (the element captured from `window.Content`).

When an app's content is hosted under a `XamlRoot` it does not own — e.g. a secondary app loaded into a collectible `AssemblyLoadContext` whose `Window.Content` is re-parented into a host's single shared `XamlRoot` — `XamlRoot.Content` resolves to the **host's** root visual. Toggling the theme then re-themes the host while the hosted app does not change. As a side effect `IsDark` reads the unchanged `RootElement.ActualTheme`, so `ThemeChanged` never fires either.

The class was internally inconsistent: it **read** (`IsDark`) and **observed** (`ActualThemeChanged`) theme from `RootElement`, but **wrote** to `RootElement.XamlRoot.Content`. Standalone apps were unaffected because there `window.Content == XamlRoot.Content`, so the bug was invisible.

## What is the new behavior?

`InternalSetThemeOnUIThread` now targets `RootElement` directly:

```diff
- var rootElement = RootElement?.XamlRoot?.Content as FrameworkElement;
- if (rootElement is null) { return false; }
+ var rootElement = RootElement as FrameworkElement;
+ if (rootElement?.XamlRoot is null) { return false; }
```

`RequestedTheme`/`ElementTheme` cascades to the subtree, so the app's own root (and only it) is themed; an ancestor host is untouched. The `XamlRoot` null-check is **retained** as the "element realized yet?" gate that drives the `Loaded`-retry in `InitializeAsync`, so the only behavioral change is *where* the theme is written when the element is realized — standalone behavior is unchanged.

Read / write / observe now all act on the same element, so `IsDark` and `ThemeChanged` work correctly in the hosted case too.

## PR Checklist

- [x] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added
- [x] Associated with an issue (#3120)

### Tests

Red/fix/green runtime test added to `Given_ThemeService.cs` (`Uno.Extensions.Core.UI.Tests`): `When_HostedUnderForeignXamlRoot_Then_ThemeAppliesToOwnRoot_NotHost`. It loads a `host → app` tree into the test window so the app root has a real `XamlRoot` whose `Content` is the host, then asserts `SetThemeAsync(Dark)` themes the app root (not the host) and that `ThemeChanged` fires. Fails on `main` (theme lands on the host), passes with the fix. The four existing `ThemeService` tests (bare, unrealized elements with a null `XamlRoot`) are unchanged — they still hit the retained gate.

> Note: these are `[RunsOnUIThread]` runtime tests with no `dotnet test` entry point; they execute in the runtime-test stages.

## Other information

- Design doc / spec is in a separate PR: #3122.
- `Uno.Toolkit.UI.SystemThemeHelper.SetApplicationTheme(XamlRoot?, ElementTheme)` has the same `root.Content.RequestedTheme` shape, but it only receives a `XamlRoot` (no element to scope to), is in a different repo, and is not in this code path — out of scope here.

### DIFF
--- a/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
+++ b/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
@@ -145,6 +145,9 @@ internal class ThemeService : IThemeService, IDisposable
 
 		if (rootElement?.XamlRoot is null)
 		{
+			// The bug this fixes (#3120) was a *silent* theme no-op, so make the unrealized case
+			// observable: without this log, "the theme just doesn't switch" leaves nothing to grep.
+			if (_logger?.IsEnabled(LogLevel.Debug) ?? false) _logger.LogDebugMessage("Theme not applied: root element is not realized yet (XamlRoot is null); InitializeAsync will retry once it is Loaded.");
 			return false;
 		}
 
@@ -157,6 +160,10 @@ internal class ThemeService : IThemeService, IDisposable
 		};
 
 		SaveDesiredTheme(theme);
+
+		// Log the apply so "applied, but no visual flip" (effective theme unchanged) is
+		// distinguishable in the field from "silently did nothing".
+		if (_logger?.IsEnabled(LogLevel.Debug) ?? false) _logger.LogDebugMessage($"Applied theme {theme} to the service's own root element.");
 
 		if (existingIsDark != IsDark)
 		{
--- a/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
+++ b/src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
@@ -133,9 +133,17 @@ internal class ThemeService : IThemeService, IDisposable
 	private bool InternalSetThemeOnUIThread(AppTheme theme)
 	{
 		var existingIsDark = IsDark;
-		var rootElement = RootElement?.XamlRoot?.Content as FrameworkElement;
 
-		if (rootElement is null)
+		// Apply the theme to the service's own root element (captured from window.Content),
+		// NOT to RootElement.XamlRoot.Content. The latter is the root visual of the XamlRoot,
+		// which differs from RootElement when the app is hosted under a XamlRoot it does not own
+		// (e.g. a secondary app whose Window.Content is re-parented into a host's shared XamlRoot,
+		// such as an app loaded into a collectible AssemblyLoadContext). Writing to XamlRoot.Content
+		// there re-themes the host instead of the app. The XamlRoot null-check is retained as the
+		// "is the element realized yet" gate that drives the Loaded-retry in InitializeAsync. See #3120.
+		var rootElement = RootElement as FrameworkElement;
+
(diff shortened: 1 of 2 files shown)


src/Uno.Extensions.Core.UI/Toolkit/ThemeService.cs
```diff
@@ -150,6 +161,10 @@ private bool InternalSetThemeOnUIThread(AppTheme theme)
 
 		SaveDesiredTheme(theme);
 
+		// Log the apply so "applied, but no visual flip" (effective theme unchanged) is
+		// distinguishable in the field from "silently did nothing".
+		if (_logger?.IsEnabled(LogLevel.Debug) ?? false) _logger.LogDebugMessage($"Applied theme {theme} to the service's own root element.");
```
nickrandolph: Don't use single line if-then statements. Always wrap in braces (apply to all cases in PR)

nickrandolph (approved):

nickrandolph (approved):

kazo0 (approved):

Issue #3130, opened 2026-07-16
[HR][Navigation] Rebase NavigationRouteUpdateHandler onto Uno's element-update pipeline (ElementMetadataUpdateHandler)


kazo0:
## Context

Tracking the Uno.Extensions-side work agreed in the HR/HD sync on 2026-07-16 (Steve Bilogan, Nick Randolph, David Rey, Matthew Rajala). Parent issue with the full investigation, repro, bundles and timeline: **unoplatform/studio.live#3079**.

Symptom there: in a generated multi-tab app, the **default tab keeps rendering its scaffolded placeholder** in the running app while Previews render the fully-built page. Non-default tabs are fine — they're instantiated lazily *after* the Hot Reload fill, so they get the new content for free. Only the page instance that was already live when the HR update landed goes stale. It's **intermittent**, which is what points at an ordering race rather than a deterministic logic bug.

### Why this lands in Uno.Extensions

Since Uno 6.7, applying HR UI updates is Uno's responsibility again: the UI-update phase removes each live instance of an updated type from the visual tree and replaces it with a fresh instance built from the latest XAML. The navigation layer has no way to block that phase — as Nick put it, *"it's still on the correct page — it's just not the latest page."*

The suspected mechanism (David's hypothesis, to be confirmed by debugging) is that the swap **does** happen, but navigation never learns about the new instance: region/route state still points at the old placeholder and re-shows it. Our route-cascade plumbing fires from a raw `[MetadataUpdateHandler]` in `NavigationRouteUpdateHandler.UpdateApplication`, which has **no ordering guarantee relative to Uno's HR UI-update phase** — the cascade can run before instances are swapped. That unordered interleaving also explains the intermittency.

## Agreed direction

Rebase `NavigationRouteUpdateHandler` from the raw `[MetadataUpdateHandler]` hook onto Uno's **element-update pipeline** — `ElementMetadataUpdateHandlerAttribute` (`Uno.UI/Helpers`), dispatched by `ElementUpdateAgent` (`Uno.UI.RemoteControl/HotReload/MetadataUpdater`).

What that buys us:

- **`BeforeVisualTreeUpdate(Type[]?)` / `AfterVisualTreeUpdate(Type[]?)`** bracket the HR UI-update phase. Cascading from `AfterVisualTreeUpdate` guarantees navigation reacts *after* instance replacement, which eliminates the ordering race outright.
[...]

PR #2293, opened 2024-05-14
docs: Updated ArcGisApp docs

vatsashah45:
GitHub Issue (If applicable): https://github.com/unoplatform/uno.samples-private/issues/146

## PR Type

What kind of change does this PR introduce?

- Documentation content changes

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tested code with current [supported SDKs](../README.md#supported)
- [x] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md). (for bug fixes / features) 
- [ ] [Unit Tests and/or UI Tests](doc/articles/working-with-the-samples-apps.md) for the changes have been added (for bug fixes / features) (if applicable)
- [ ] [Wasm UI Tests](doc/articles/working-with-the-samples-apps.md#running-the-webassembly-ui-tests-snapshots) are not showing unexpected any differences. Validate PR `Screenshots Compare Test Run` results.
- [x] Contains **NO** breaking changes
- [ ] Updated the [Release Notes](https://github.com/unoplatform/uno/tree/master/doc/ReleaseNotes)
- [x] Associated with an issue (GitHub or internal)

## Other information

Internal Issue (If applicable):

PR #3131, opened 2026-07-17
fix: re-hook FrameNavigator after external hot-reload content swap

rajamatt:
GitHub Issue (If applicable): related to #3130

## PR Type

- Bugfix

In a multi-tab app whose pages are filled by an external Hot Reload (agent/IDE) while the app's view has not had a layout pass, the default tab keeps showing its scaffolded placeholder (#3130). The investigation showed the live default page is **never replaced**: it was created while its host panel could not lay out, so it exists only as `Frame.Content` — never as a materialized visual child — and Uno's HR visual-tree walk cannot see it. Navigation then declines to refresh it by design (the keep-active-instance cascade skip, and `no segments to navigate` on re-selection), so the stale placeholder is what appears once the view finally lays out.

Additionally, even when the page instance is replaced outside navigation, `FrameNavigator` keeps tracking the dead instance (no `Frame.Navigated` is raised), and the new instance carries a copied, stale `DataContext` — so bound content renders empty.

## What is the new behavior?

Together with the Uno-side fix (Uno's `FrameElementMetadataUpdateHandler` now re-creates unmaterialized frame content — PR link TBD), navigation now fully recovers:

- New `NavigationFrameContentUpdateHandler` (element-update pipeline, `AfterVisualTreeUpdate`): after each HR visual-tree phase, every `FrameNavigator` in the live region trees re-hooks onto its frame's actual content. Empty-type-list cycles are ignored.
- `FrameNavigator.RehookCurrentViewAfterHotReload`: updates `CurrentView`, re-applies the region name, re-parents nested regions, and rebuilds the view model through the standard pipeline (`InitializeCurrentView` with `refresh: true` — required because an in-place EnC update keeps the type identity, which the default "wrong VM type" check treats as still valid). Bindings then target a view model built from the updated types instead of the stale copied `DataContext`.
- Deterministic regression test: `Given_HotReload.When_PageXamlUpdatedWhileUnmaterialized_Then_RevealShowsUpdatedContent` reproduces the stranded-page state on Skia desktop by collapsing the host panel before populating it. Red on current Uno packages; green once the Uno-side fix ships in the pinned Uno version. `HotReload.Spec.md` documents the scenario and harness constraints.
- No regressions: full navigation HR suite (6) and TabBar HR suite (23) pass with both fixes applied.

## Other information

The new runtime test is expected to fail until the Uno-side fix (unoplatform/uno/pull/23804) is available in the Uno version this repo pins — sequence the two PRs accordingly (land the Uno fix first, or merge this one together with the version bump that includes it). The temporary `[NAV-HR-DIAG]` diagnostic logging added earlier on this branch is kept for now to help validate the fix in real runs; it can be removed or downgraded to Debug in a follow-up.


src/Uno.Extensions.Navigation.UI/Navigators/FrameNavigator.cs
```diff
@@ -652,6 +685,74 @@ protected override void UpdateRoute(Route? route)
...
+			_ = RefreshViewModelAfterHotReloadAsync(route, mapping);
+		}
+	}
+
+	private async Task RefreshViewModelAfterHotReloadAsync(Route route, RouteInfo mapping)
+	{
+		try
+		{
+			var request = new NavigationRequest(this, route);
+			await InitializeCurrentView(request, route, mapping, refresh: true);
```
autocarl: [High] `refresh: true` discards the copied DataContext on every frame-content replacement, even when the HR delta only changed XAML. The linked Uno frame fix deliberately copies a locally-set DataContext to the replacement page; this call then always resolves a new transient VM because the type-identity guard is bypassed. A plain XAML edit therefore loses unsaved VM state, subscriptions, and other instance state. The new stranded-content test registers no VM, so it cannot catch this regression. Please preserve the existing DataContext for XAML-only updates and force recreation only when the updated types show that the mapped VM actually needs it.

src/Uno.Extensions.Navigation.UI/Navigators/FrameNavigator.cs
```diff
@@ -652,6 +685,74 @@ protected override void UpdateRoute(Route? route)
...
+
+		// The replaced instance carries the stale placeholder's DataContext (Uno's frame
+		// element-update handler copies it during the swap), so every binding targets a
+		// view model built from the pre-update types and renders empty. Rebuild the view
+		// model through the standard pipeline; refresh:true is required because an
+		// in-place (EnC) update keeps the type identity, which the default "wrong type"
+		// check would treat as still valid.
+		if (Route is { Base.Length: > 0 } route && Resolver.FindByPath(route.Base) is { } mapping)
+		{
+			_ = RefreshViewModelAfterHotReloadAsync(route, mapping);
```
autocarl: [High] This starts a navigation-like VM initialization outside the awaited navigation pipeline. `InitializeCurrentView` reaches `CreateViewModel`, whose `Task.Run` mutates the region scope's shared `NavigationDataProvider.Parameters` and `IScopedInstanceRepository` dictionary. If the user navigates (or another HR cycle arrives) before this task completes, both operations can write the same scope concurrently, so the wrong `NavigationRequest`/route data can be injected or the non-thread-safe dictionary can fail. Please serialize this refresh with navigation, or use an isolated scope plus cancellation/generation checks before applying the result.

src/Uno.Extensions.Navigation.UI/NavigationHotReloadDiagnostics.cs
```diff
@@ -0,0 +1,168 @@
...
+		var indent = new string(' ', depth * 2);
+		var navigator = region.Navigator();
+		var view = region.View;
+		var viewInfo = view is null
+			? "<null>"
+			: $"{Id(view)} loaded={view.IsLoaded} vis={view.Visibility} opacity={view.Opacity}";
+
+		Region.Logger.LogWarningMessage(
+			$"{Tag}   {indent}region name='{region.Name ?? string.Empty}' navigator={Id(navigator)} " +
+			$"route='{navigator?.Route?.ToString() ?? "<null>"}' view={viewInfo}");
```
autocarl: [High] `Route.ToString()` appends `Route.Query()`, and `Query()` serializes every non-empty `Route.Data` key/value. This warning-level region dump can therefore write access tokens, identifiers, and other PII into normal logs and feedback bundles during Hot Reload. `Navigator.cs` already documents this exact risk and logs only `Route.Base`. Please log only structural route fields (or redact all values) here as well.

autocarl (changes requested):
Static review only; I did not run local builds or tests. I found three material issues on this head: XAML-only frame swaps unconditionally recreate the view model, the fire-and-forget refresh can overlap normal navigation through shared scoped state, and the warning-level diagnostics serialize Route.Data. The hosted Hot Reload failure is the expected missing Uno dependency, but these findings are independent of that dependency.

src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
```diff
@@ -58,9 +65,13 @@ internal static void Unregister(NavigationRouteContext context)
...
 	{
-		if (Region.Logger.IsEnabled(LogLevel.Information))
+		if (Region.Logger.IsEnabled(LogLevel.Warning))
 		{
-			Region.Logger.LogInformationMessage($"Hot-reload UpdateApplication received {updatedTypes?.Length ?? 0} updated type(s); refreshing {_contexts.Count} navigation context(s)");
+			// [NAV-HR-DIAG] #3130: log the actual type names (not just the count) plus the thread,
+			// so ordering relative to Uno's HR UI-update phase (ElementUpdateAgent callbacks)
+			// can be reconstructed from the logs.
+			var typeNames = updatedTypes is null ? "<null>" : string.Join(", ", updatedTypes.Select(t => t.FullName ?? t.Name));
+			Region.Logger.LogWarningMessage($"[NAV-HR-DIAG] UpdateApplication (raw MetadataUpdateHandler) types=[{typeNames}]; refreshing {_contexts.Count} navigation context(s) (thread={Environment.CurrentManagedThreadId})");
```
kazo0: This is debugging logs that we don't need to actually commit, no?

src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
```diff
@@ -279,13 +298,21 @@ private static void ScheduleCascade(IRegion root, RouteResolver resolver)
 			return;
 		}
 
-		if (Region.Logger.IsEnabled(LogLevel.Debug))
+		if (Region.Logger.IsEnabled(LogLevel.Warning))
 		{
-			Region.Logger.LogDebugMessage($"Scheduling cascade walk on dispatcher (root region: '{root.Name ?? string.Empty}', children: {root.Children.Count})");
+			Region.Logger.LogWarningMessage($"[NAV-HR-DIAG] Scheduling cascade walk on dispatcher (root region: '{root.Name ?? string.Empty}', children: {root.Children.Count}, thread={Environment.CurrentManagedThreadId})");
```
kazo0: Same comment

src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
```diff
@@ -257,11 +272,15 @@ internal static void ScheduleCascadeForAllContextsIfRouteRelevant(Type[]? update
...
+					{
+						Region.Logger.LogWarningMessage("[NAV-HR-DIAG] Visibility-restore cascade WILL be scheduled (RestoreState path).");
+					}
 					ScheduleCascade(root, resolver);
 				}
-				else if (Region.Logger.IsEnabled(LogLevel.Debug))
+				else if (Region.Logger.IsEnabled(LogLevel.Warning))
 				{
-					Region.Logger.LogDebugMessage("Hot-reload visibility-restore cascade skipped: none of the updated types are registered navigation routes.");
+					Region.Logger.LogWarningMessage("[NAV-HR-DIAG] Hot-reload visibility-restore cascade skipped: none of the updated types are registered navigation routes.");
```
kazo0: Same comment, remove?

src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs
```diff
@@ -279,13 +298,21 @@ private static void ScheduleCascade(IRegion root, RouteResolver resolver)
...
+			Region.Logger.LogWarningMessage($"[NAV-HR-DIAG] Scheduling cascade walk on dispatcher (root region: '{root.Name ?? string.Empty}', children: {root.Children.Count}, thread={Environment.CurrentManagedThreadId})");
 		}
 
 		dispatcher.TryEnqueue(() =>
 		{
+			// [NAV-HR-DIAG] #3130: timestamp when the deferred cascade actually runs — the
+			// ordering race is between this callback and Uno's HR visual-tree swap.
+			if (Region.Logger.IsEnabled(LogLevel.Warning))
+			{
+				Region.Logger.LogWarningMessage($"[NAV-HR-DIAG] Deferred cascade RUNNING on dispatcher (root region: '{root.Name ?? string.Empty}', thread={Environment.CurrentManagedThreadId})");
```
kazo0: same

src/Uno.Extensions.Navigation.UI.Tests/Given_HotReload.cs
```diff
@@ -347,6 +347,164 @@ public async Task When_HRCascadeAfterUserSelection_Then_ActiveRegionPreserved(Ca
...
+	/// <c>VisualTreeHelper</c> children only, so a XAML hot reload of that page's type
+	/// replaces nothing, and navigation keeps the stale instance (the keep-active-instance
+	/// cascade skip + "no segments to navigate" both decline to refresh). Revealing the
+	/// panel afterwards shows the pre-HR placeholder — the "default tab renders its
+	/// scaffolded placeholder" symptom from the original report.
+	/// </summary>
+	[TestMethod]
+	[RunsOnUIThread]
+	public async Task When_PageXamlUpdatedWhileUnmaterialized_Then_RevealShowsUpdatedContent(CancellationToken ct)
+	{
```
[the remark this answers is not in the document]
rajamatt: Yes, intended. Will update when uno version is available

src/Uno.Extensions.Navigation.UI/NavigationFrameContentUpdateHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+		{
+			frameNavigator.RehookCurrentViewAfterHotReload(updatedTypes);
+		}
+
+		foreach (var child in region.Children.ToArray())
+		{
+			RehookReplacedFrameContent(child, updatedTypes);
+		}
+	}
+}
```
kazo0, line 56: For this work here, don't we need to disable what was previously done as part of the original MetadataUpdateHandler work here? https://github.com/unoplatform/uno.extensions/blob/88cc6659a82bf4b82a948ea877f860dc30ec5f75/src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs#L81

Where are are still re-cascading and rebuilding all of the routes?

Claude's finding:

> It doesn't do what #3130 agreed to. The acceptance criteria say "the IsDefault cascade runs strictly after Uno's visual-tree update phase; no reliance on TryEnqueue" and "remove the now-redundant defensive skip." The PR does neither: the cascade still fires from the raw [MetadataUpdateHandler] with the TryEnqueue workaround, the #2293 keep-active-instance skip survives, and a third handler is added alongside. So uno.extensions now has raw-hook cascade + visibility handler + frame-content handler, all partially overlapping in responsibility. If the rebase is deferred, #3130 should be re-scoped explicitly; right now the PR claims "related to #3130" while leaving its two central tasks unchecked.

src/Uno.Extensions.Navigation.UI/Navigators/FrameNavigator.cs
```diff
@@ -652,6 +656,142 @@ protected override void UpdateRoute(Route? route)
...
+	/// (e.g. for a page that was never materialized in the visual tree — see
+	/// uno.extensions#3130); that path raises no <c>Frame.Navigated</c>, so
+	/// <see cref="CurrentView"/> would keep pointing at the dead instance and nested
+	/// regions would stay attached to it.
+	/// </summary>
+	/// <param name="updatedTypes">
+	/// The types applied by the hot-reload delta; used to decide whether the mapped view
+	/// model must be rebuilt. <see langword="null"/> means the delta's types are unknown.
+	/// </param>
+	internal void RehookCurrentViewAfterHotReload(Type[]? updatedTypes)
```
kazo0, line 688: Claude's question seems valid:

> RehookCurrentViewAfterHotReload fires the VM rebuild as fire-and-forget from inside Uno's UI-update phase, while the raw-hook cascade may also be queued for the same cycle — the interplay between the rehook's InitializeCurrentView(refresh: true) and a concurrently scheduled cascade navigation on the same region isn't reasoned about anywhere in the PR.

rajamatt: Added some comments to the method, but I think the two can't conflict. Basically the re-hook never triggers navigation on its own, and if a real navigation happens before the queued VM refresh runs, the refresh is skipped

src/Uno.Extensions.Navigation.UI.Tests/Given_HotReload.cs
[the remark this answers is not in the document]
rajamatt: Uno fix merged

src/Uno.Extensions.Navigation.UI/Navigators/FrameNavigator.cs
```diff
@@ -652,6 +656,152 @@ protected override void UpdateRoute(Route? route)
...
+					// Re-check before applying: CreateViewModel resumes off the dispatcher,
+					// and a navigation that completed meanwhile must win.
+					if (token != _hotReloadRefreshToken || !ReferenceEquals(Control?.Content, view))
+					{
+						return;
+					}
+				}
+
+				await view.InjectServicesAndSetDataContextAsync(services, navigator, viewModel);
+			});
```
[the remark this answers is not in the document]
rajamatt: The `await` captures the UI thread's synchronization context, so everything after `CreateViewModel` resumes back on the dispatcher

src/Uno.Extensions.Navigation.UI/NavigationFrameContentUpdateHandler.cs
rajamatt: I got this:

> The old handler still has to rebuild the routes (C#-only changes never reach Uno's element-update pipeline), and the new handler only re-hooks the frame content and never navigates, so they don't overlap.

src/Uno.Extensions.Navigation.UI/NavigationFrameContentUpdateHandler.cs
```diff
@@ -0,0 +1,56 @@
+using System.Reflection.Metadata;
+using Uno.Extensions.Navigation.Navigators;
+using Uno.Extensions.Navigation.Regions;
+
+[assembly: ElementMetadataUpdateHandlerAttribute(typeof(FrameworkElement), typeof(Uno.Extensions.Navigation.UI.NavigationFrameContentUpdateHandler))]
```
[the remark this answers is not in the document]
rajamatt: The file compiles fine

src/Uno.Extensions.Navigation.UI/NavigationFrameContentUpdateHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+		// means "types unknown" and must still re-hook: the mismatch check below is what
+		// decides whether anything actually changed.
+		if (updatedTypes is { Length: 0 })
+		{
+			return;
+		}
+
+		foreach (var root in NavigationRouteUpdateHandler.ActiveContexts
+			.Select(static ctx => ctx.RootRegion)
+			.OfType<IRegion>())
```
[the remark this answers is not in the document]
rajamatt: Same as above, compiles fine

src/Uno.Extensions.Navigation.UI/Navigators/FrameNavigator.cs
```diff
@@ -652,6 +656,152 @@ protected override void UpdateRoute(Route? route)
...
+				// frame element-update handler copies it during the swap). Rebuild the view
+				// model only when it is missing, of the wrong type, or its type was part of
+				// the delta — an in-place (EnC) update keeps the type identity, which the
+				// "wrong type" check alone would treat as still valid. For a XAML-only delta
+				// the copied DataContext IS the live view model and is preserved, along with
+				// any un-persisted state it holds.
+				var viewModel = view.DataContext;
+				if (viewModelUpdated || viewModel is null || viewModel.GetType() != mapping.ViewModel)
+				{
+					viewModel = await CreateViewModel(services, new NavigationRequest(this, route), route, mapping);
```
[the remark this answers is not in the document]
rajamatt: When `mapping.ViewModel` is null, `CreateViewModel` returns null right away and `InjectServicesAndSetDataContextAsync` skips null VMs

kazo0 (approved):

kazo0:
/unobot prepare-release --commit-message-pattern "ci: Set version to '{0}'"

Issue #3072, opened 2026-04-30
[P1][HR][Gaps][Navigation] Adding/removing a RouteMap at runtime has no effect


kazo0:
## Source
Sub-issue of unoplatform/private#926. Tracked from the [HR/HD test matrix](https://github.com/unoplatform/private/issues/926#issuecomment-3371906256).

## Current behavior
Adding or removing a `RouteMap` at runtime via Hot Reload has no effect — the new/changed routes are not picked up by the navigator until the app is restarted.

## Expected behavior
Edits to the route registration code (`App.xaml.host.cs` or wherever `RouteMaps` are registered) should apply via Hot Reload so newly added routes can be navigated to, and removed routes should stop resolving, without requiring an app restart.

## How to reproduce
1. Run an Uno.Extensions navigation sample (e.g. Chefs).
2. While the app is running, add a new `RouteMap` entry to `RegisterRoutes` (e.g. `new RouteMap("NewRoute", View: views.FindByViewModel<NewModel>())`).
3. Apply Hot Reload.
4. Try to navigate to the new route via XAML or code-behind — it does not resolve.
5. Repeat with removal — removed routes still resolve until restart.

## Goal
- Make `RouteMaps` registration HR-aware so route additions/removals/edits apply at runtime.
- Once fixed, add a regression test in the Hot Reload testharness that mutates `RegisterRoutes` and asserts the new/removed route is honored.

## Priority
Failing scenario — high priority for 6.5 / .NET Conf 2025 demo readiness.

PR #3146, opened 2026-08-20
fix: re-instantiate the active route's view model on hot reload

### LINKED ISSUE #3142: [HR][Navigation] CascadeNewDefaultsFromRoot suppresses the refresh of the active route, so the displayed page is never re-instantiated
## Current behavior

When a hot-reload delta updates a type that backs the **currently displayed** route, the page and its
view model are never re-instantiated, so the change is invisible until a full rebuild.

`UpdateApplication` runs, the route table rebuilds, `ShouldCascadeForUpdatedTypes` correctly decides the
change is navigation-relevant — and then `CascadeNewDefaultsFromRoot` suppresses the navigation because
the target region is *already* on that route. Nothing else forces a re-navigation.

This is most visible for edits that only affect a constructor or property initializer, because a
metadata update never re-runs constructors on live instances:

```csharp
public ISeries[] DeploymentSeries { get; } = new ISeries[]
{
    new ColumnSeries<double> { Values = new double[] { 8, 14, 11, 22 } },  // hot-reload edit -> LineSeries<double>
};
```

That class of edit *requires* the page to be re-created to become visible. When the page is the active
route, it never is — so the effect is that the one page guaranteed not to refresh is the page the user
is looking at.

## Expected behavior

If an updated type is the view or view model of a region's **active** route, that region should be
re-navigated (re-instantiated) rather than skipped.

## How to reproduce

1. App using `Uno.Extensions.Navigation` with a `NavigationView` shell and nested routes, e.g.

   ```csharp
   views.Register(
       new ViewMap<MainPage, MainModel>(),
       new ViewMap<OverviewPage, OverviewModel>(),
       new ViewMap<PerformancePage, PerformanceModel>());

   routes.Register(
       new RouteMap("Main", View: views.FindByViewModel<MainModel>(),
           Nested:
           [
               new RouteMap("Overview", View: views.FindByView<OverviewPage>(), IsDefault: true),
               new RouteMap("Performance", View: views.FindByView<PerformancePage>()),
           ]));
   ```

2. Run the app and stay on the default `Overview` page.
3. Edit a property initializer on `OverviewModel` (a value that is only computed in the constructor).
4. Apply hot reload.
5. The UI does not change. The log shows the cascade being skipped.

Step 3 is the important part — a method-body edit or a XAML edit takes a different path and is not
affected.

## Logs

```
info: Uno.Extensions.Navigation.Regions.NavigationRegion[0]
      UpdateApplication - Hot-reload UpdateApplication received 1 updated type(s); refreshing 1 navigation context(s)
info: Uno.Extensions.Navigation.MappedRouteResolver[0]
      LogResolverState - RouteResolver rebuilt: First='Main', TopLevelRoutes=2, TotalMappings=6
info: Uno.Extensions.Navigation.MappedRouteResolver[0]
      InternalDefaultMapping - For better performance (avoid reflection), create mapping for path 'OverviewModel', view 'OverviewModel', view model ''
warn: Uno.Extensions.Navigation.MappedRouteResolver[0]
      TypeFromPath - Navigation failed: Could not resolve type for path 'OverviewModel'.
dbug: Uno.Extensions.Navigation.MappedRouteResolver[0]
      InternalDefaultMapping - Created default mapping - Path 'OverviewModel'
dbug: Uno.Extensions.Navigation.Regions.NavigationRegion[0]
      ScheduleCascade - Scheduling cascade walk on dispatcher (root region: '', children: 1)
dbug: Uno.Extensions.Navigation.Regions.NavigationRegion[0]
      CascadeNewDefaultsFromRoot - Cascade skip on region '': IsDefault 'Overview' suppressed because descendant region 'Overview' already has active route 'Overview' (matches nested path of 'Main')
```

The `TypeFromPath` warning on line 4 is a red herring — `InternalDefaultMapping` recovers on the next
line ("Created default mapping"). It is noise, not the failure. (Separately: it would be nice for that
advisory not to log at `warn` when the fallback succeeds.)

The last line is the defect.

## Root cause

In `src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs`:

- `UpdateApplication(Type[] updatedTypes)` → `RebuildRoutes` → `ShouldCascadeForUpdatedTypes(updatedTypes, resolver)`
  returns `true` when at least one updated type is registered as a view or view model. That check exists
  specifically to *avoid* re-mounting pages on XAML-only edits — so it has already concluded this edit
  should re-mount something.
- `ScheduleCascade` → `CascadeNewDefaultsFromRoot(IRegion region, RouteResolver resolver)` calls
  `FindActiveDescendantNestedRoute`. When a descendant region's active route already matches one of the
  nested routes, the `IsDefault` cascade is skipped to preserve that selection.
- `RetryPendingFailedRequestsFromRoot` only re-invokes `RetryPendingFailedRequestAsync` on navigators
  that hold a *pending failed* request. A region that navigated successfully at startup has none.
- Net: there is no path that forces re-instantiation of a region whose active route corresponds to an
  updated type. Note that `CascadeNewDefaultsFromRoot` never receives `updatedTypes` at all.

The suppression itself is intentional and correct for the case it was added for — from
`Given_HotReload.cs`:

> the fix ensures that walk respects RegionTwo's active selection and does NOT re-issue an IsDefault
> navigation back to RegionOne

The problem is that two different situations share one branch:

| Situation | Correct action | Today |
|---|---|---|
| Cascade wants a *different* default than the region's active route | suppress — don't move the user | suppress ✅ |
| The active route's **own** view/view model was just updated | re-navigate to re-instantiate it | suppress ❌ |

## Suggested fix

Thread `updatedTypes` through to the conflict check in `CascadeNewDefaultsFromRoot`, and when the
conflicting region's active route resolves to a view or view model in `updatedTypes`, force a
re-navigation of that same route instead of suppressing. `ShouldCascadeForUpdatedTypes` already performs
the "is this type registered as a view/view-model" lookup, so no new plumbing is needed — only the
condition changes.

Regression test: extend `src/Uno.Extensions.Navigation.UI.Tests/Given_HotReload.cs` to navigate to a
region, edit the **active** region's view model, and assert the page/view model is re-instantiated —
alongside the existing assertion that an unrelated `IsDefault` cascade does not yank the selection back.

## Environment

Nuget Package (s): Uno.Extensions.Navigation, Uno.Extensions.Navigation.UI

Package Version(s): as shipped with Uno SDK 6.7.16

Affected platform(s):

- [x] WebAssembly
- [x] Windows
- [x] Skia (Desktop)

The affected code path is in `Uno.Extensions.Navigation.UI` and is not platform specific.

Code references are against `main` @ `3f5d79e4c4fad547bd651c4766f9d753d637330f`.

## Related

- #3072 — adding/removing a `RouteMap` at runtime has no effect. Same area (hot reload vs. the route
  table), different trigger: that one is about the route *table* not updating, this one is about a
  route that resolves correctly but is never re-navigated.

rajamatt:
GitHub Issue (If applicable): closes #3142

## PR Type

- Bugfix

A C# hot-reload delta that updates the view model of the **currently displayed** route never
re-instantiates that view model, so edits that only run at construction time (constructor bodies,
property initializers) stay invisible until a full rebuild — the one page guaranteed not to
refresh is the page the user is looking at.

Every existing HR path declines to handle it:

- `NavigationRouteUpdateHandler.UpdateApplication` rebuilds the route table and correctly
  classifies the delta as navigation-relevant, but `CascadeNewDefaultsFromRoot` suppresses the
  walk because the region is *already* on that route (the suppression itself is correct — it
  preserves the user's selection).
- `FrameNavigator.RehookCurrentViewAfterHotReload` only reacts when the frame's content element
  was replaced; a metadata-only edit to a plain C# view-model class replaces no element.
- `RetryPendingFailedRequestsFromRoot` only re-issues *failed* requests.

Re-issuing the same route is also a no-op end to end: `FrameNavigator.NavigateForwardAsync`
computes 0 forward segments for the route it is already on, and its HR bypass only fires when
`SourcePageType` differs from the mapping's `RenderView` — which it doesn't for an in-place
(EnC) update.

## What is the new behavior?

When a hot-reload delta updates the view model mapped to a region's **active** route, that view
model is re-created in place and rebound — no navigation is issued, so the user's selection
cannot move.

Implementation (two orthogonal walks instead of entangling `updatedTypes` into the cascade's
conflict check — `CascadeNewDefaultsFromRoot` is unchanged):

- `NavigationRouteUpdateHandler.ScheduleCascade` now receives `updatedTypes`.
  `CollectUpdatedViewModels` maps each updated type to the view-model type the just-rebuilt route
  table associates with it, resolved through `RouteResolver.FindByViewModel` so the MVUX case
  works (`MappedRouteResolver` translates a model type — what the delta contains — to its
  generated bindable view model — what `RouteInfo.ViewModel` holds).
- `RefreshUpdatedActiveRoutesFromRoot` walks the live region tree; every region whose navigator's
  active `Route.Base` resolves to a mapping whose `ViewModel` is in that set gets the new
  `ControlNavigator.RefreshActiveRouteViewModelAsync()`, which re-runs the standard
  `InitializeCurrentView(refresh: true)` pipeline (`CreateViewModel` +
  `InjectServicesAndSetDataContextAsync`) for the route the navigator is already on.
- **View types deliberately do not trigger the refresh**: updated views are owned by Uno's
  element-update walk (plus the frame-content re-hook), and rebuilding view models on a view-only
  delta would discard un-persisted view-model state on every page edit — the behavior
  `Given_FrameContentRehook.When_XamlOnlyDelta_Then_CopiedViewModelIsPreserved` pins.
- `updatedTypes == null` (unknown delta) refreshes nothing — conservative, mirrors the existing
  x:Uid-revert guard on the cascade.

Known limitation (documented in `specs/009-hr-active-route-vm-refresh/spec.md`): a view model
replaced via `CreateNewOnMetadataUpdate` (rude edit / some partial-class deltas) resolves to the
old type in the rebuilt route table and is not matched — same pre-existing limitation as the
route table itself (#3072). The in-place EnC case (type identity preserved) is fully covered.

## Other information

**Tests added** (red/fix/green — the repro tests fail on `main`):

- `Given_ActiveRouteVmRefresh` (UI test, no HR harness — same technique as
  `Given_FrameContentRehook`): drives the internal HR entry point with a synthetic delta against
  a live navigation tree. Asserts the VM is rebuilt in place (view instance untouched) for a
  VM delta, and preserved for a view-only delta. Verified red before the fix (timeout — nothing
  re-instantiates the VM) and green after, on Skia desktop.
- `Given_HotReload.When_ActiveRouteViewModelUpdated_Then_ActiveRegionVmReinstantiated` (runtime
  test, real HR delta, secondary app): edits a constructor-seeded property initializer on the
  active route's own VM file while sitting on a **non-default** visibility-region route; asserts
  the VM is re-instantiated with the post-HR seed AND the selection is not yanked back to the
  `IsDefault` route. Runs in the hot-reload runtime-test CI stage.
- `Given_NavigationRouteUpdateHandler`: 5 unit tests for `CollectUpdatedViewModels`
  (null / unregistered / registered VM / registered view-only / MVUX mapped-model).
- Existing guard suites verified green: `Given_FrameContentRehook` (state preservation
  unchanged), `Given_NavigationRouteUpdateHandler`, neighboring navigation suites.

Also validated end-to-end in a downstream consumer: with this assembly, editing a chart series
property initializer on the displayed page's model (`ColumnSeries<double>` →
`LineSeries<double>`) renders live on the active page in one hot-reload pass — previously it
required breaking hot reload badly enough to force a full rebuild.

Design doc: `specs/009-hr-active-route-vm-refresh/spec.md` (+ scenario entry in
`src/Uno.Extensions.Navigation.UI.Tests/HotReload.Spec.md`).

### DIFF
--- a/specs/009-hr-active-route-vm-refresh/progress.md
+++ b/specs/009-hr-active-route-vm-refresh/progress.md
@@ -1,0 +1,45 @@
+# Progress — HR refresh of the active route's view model (#3142)
+
+Branch: `dev/mara/fix-routes-hr` (no commit/push in this session — per request)
+
+## Plan
+
+- [x] Read issue #3142, the downstream consumer report that surfaced it, and the existing HR machinery
+      (`NavigationRouteUpdateHandler`, `FrameNavigator` rehook, `Given_FrameContentRehook`,
+      spec 003, `HotReload.Spec.md`)
+- [x] Write spec (`spec.md` in this folder)
+- [x] RED: harness-free UI test `Given_ActiveRouteVmRefresh` (VM delta → rebuilt; view-only delta →
+      preserved) — fails on current code
+- [x] RED: runtime HR test `Given_HotReload.When_ActiveRouteViewModelUpdated_Then_ActiveRegionVmReinstantiated`
+      + ctor-seeded property on `HotReloadRegionVm`
+- [x] FIX: `ControlNavigator.RefreshActiveRouteViewModelAsync` (internal virtual + generic override)
+- [x] FIX: `NavigationRouteUpdateHandler` — thread `updatedTypes` into `ScheduleCascade`, add
+      `CollectUpdatedViewModels` + `RefreshUpdatedActiveRoutesFromRoot`
+- [x] Unit tests for `CollectUpdatedViewModels` in `Given_NavigationRouteUpdateHandler`
+- [x] GREEN: build + run the new tests (and the neighboring HR/rehook tests) locally
+- [x] Update living spec `src/Uno.Extensions.Navigation.UI.Tests/HotReload.Spec.md`
+- [x] Baseline-diff the neighboring navigation suites (stash fix → rebuild → rerun) to attribute
+      any failures
+
+## Review
+
+- Verified locally on Skia desktop (net9.0-desktop runtime-test host):
+  - `Given_ActiveRouteVmRefresh` — RED before fix (`When_ActiveRouteViewModelInDelta_...` timed
+    out: nothing re-instantiated the VM), GREEN after; the view-only control test passed in both
+    states.
+  - `Given_NavigationRouteUpdateHandler` — 11/11 green (incl. 5 new `CollectUpdatedViewModels`
+    cases: null, unregistered, registered VM, registered view-only, MVUX mapped-model).
+  - `Given_FrameContentRehook` — 4/4 green (state-preservation behavior unchanged).
+  - Release build of `Uno.Extensions.Navigation.WinUI.csproj`: 0 errors.
+  - `Given_HotReload` secondary-app tests require the dev-server HR harness (not available in
+    this environment — the RC connection has no listener); the new test compiles and follows the
+    existing harness patterns and runs in CI's `stage-build-runtimetests-skia-hotreload.yml`
+    (filter `_HotReload` matches `Given_HotReload`).
+- Neighboring suites: `Given_NavigationBoundary`, `Given_Navigation_VmCreationFailure`,
+  `Given_TabNavigation`, `Given_NestedNavReloadRecovery` green. `Given_ChainedGetDataAsync` (8),
+  `Given_NavigatorStartup` (1), `Given_Region` (1), `Given_RouteNotifier` (1) failed identically
+  on the UNMODIFIED baseline (stash → rebuild → rerun) — pre-existing/environmental, not caused
+  by this change; none of them exercise any HR entry point.
+- Design decision worth keeping in mind: the refresh triggers on **view-model** membership only —
+  view deltas are owned by Uno's element walk, and including them would regress the
+  state-preservation behavior pinned by `Given_FrameContentRehook`.
--- a/specs/009-hr-active-route-vm-refresh/spec.md
+++ b/specs/009-hr-active-route-vm-refresh/spec.md
@@ -1,0 +1,105 @@
+# Hot-Reload Refresh of the Active Route's View Model
+
+**Status:** Implemented (branch `dev/mara/fix-routes-hr`)
+**Issue:** [#3142](https://github.com/unoplatform/uno.extensions/issues/3142) — `[HR][Navigation]
+CascadeNewDefaultsFromRoot suppresses the refresh of the active route, so the displayed page is
+never re-instantiated` (surfaced by a downstream hosted-app consumer)
+**Affects:** `Uno.Extensions.Navigation.UI`
+**Files touched:**
+
+- `src/Uno.Extensions.Navigation.UI/NavigationRouteUpdateHandler.cs`
+- `src/Uno.Extensions.Navigation.UI/Navigators/ControlNavigator.cs`
+
+## Problem
+
+A C# hot-reload delta that updates the view model of the **currently displayed** route never
+re-instantiates that view model, so edits that only run at construction time (constructor bodies,
+property initializers) stay invisible until a full rebuild:
+
+```csharp
+public ISeries[] DeploymentSeries { get; } = new ISeries[]
+{
+    new ColumnSeries<double> { ... },  // HR edit -> LineSeries<double>: invisible on the active page
+};
+```
+
+Why every existing HR path declines to handle it:
+
+- `NavigationRouteUpdateHandler.UpdateApplication` rebuilds the route table and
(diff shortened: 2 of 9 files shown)


kazo0 (approved):

kazo0:
@Mergifyio backport release/stable/7.3

PR #3148, opened 2026-08-21
[POC] feat(reactive): mock data sources for FeedView (POCO / JSON envelopes)

nickrandolph:
GitHub Issue (If applicable): none — **this is a proof of concept**; no issue exists yet. If the approach is validated, an issue will be created before this graduates from POC.

> [!IMPORTANT]
> **🧪 PROOF OF CONCEPT** — opened as a draft to validate the approach and the envelope convention. Not intended to merge as-is.

## PR Type

- Feature (POC)

## Initial requirements (as provided)

The goal is to make it easy to supply a mock DataContext to a page that uses a `FeedView`, using either a set of raw POCO objects, or an Expando object created from JSON:

- Currently, if a plain object is passed into a `FeedView` it's just discarded. The simple solution is to wrap the object as a feed with an initial value.
- However, that doesn't help mock different states. As part of creating the feed from an initial value, the object should be checked for `Data`, `InProgress`/`IsProgress` and `Exception` properties:
  - If a `Data` property exists, its value is used as the value of the feed in the **Value** state — unless the value is `null`, in which case the feed is in the **Empty (None)** state.
  - If a progress property exists and is `true`, the feed is in the **in-progress** state.
  - If an `Exception` property exists and is not `null`, the feed is in the **Error** state.
- A demo page in the Playground app showing both POCO objects and raw JSON data (DataContext set from JSON in XAML).

`FeedView.Source` silently discards any value that is not an `ISignal<IMessage>` (`IFeed<T>` / `IState<T>`): the control stays in its loading state and renders nothing.

## What is the new behavior?

- A non-feed `Source` is coerced into a single-message feed (internal `MockFeed`) instead of being discarded.
- A mock **envelope** — an object whose public members are limited to `Data`, `Progress` (aliases `IsProgress`/`InProgress`), `Error` (alias `Exception`), matched case-insensitively — drives the FeedView states: Value / None / Indeterminate / Error. A string error value is wrapped as the exception message (so JSON can mock errors).
- Works with POCOs/anonymous objects (reflection) and `IDictionary<string, object?>`/`ExpandoObject` (trim-safe, no reflection) — the latter is how JSON-shaped data flows in.
- An envelope whose `Data` is itself a feed passes that feed through untouched; objects with unrecognized members are rendered whole as the feed value (no false envelope detection).
- Mocked sources honor refresh requests, so `FeedView.Refresh` completes instead of hanging `IsExecuting`.
- New public API: `Feed.Value<T>(T)` — a constant single-message feed (the only public surface addition; the coercion itself is internal).
- Playground: new **FeedView Mock Data** page showing seven scenarios (POCO value, envelope progress/error/none, JSON envelope value/error, JSON plain object), each with its XAML displayed alongside the live control. Verified running on `net9.0-desktop`.

Design details and decisions are captured in `specs/009-feedview-mock-source/spec.md` (+ `progress.md`).

### Notes / known issues surfaced while building the POC

- **Pre-existing on `main`:** Playground desktop started with a blank window — initial navigation resolves no route ("Route '' has 19 nested route(s) but none marked IsDefault") despite `initialViewModel: typeof(HomeViewModel)`. Worked around here by marking the `Home` route `IsDefault: true` (isolated commit, easy to drop). The resolver behavior deserves its own issue.
- The Playground `JsonDataContext` helper deliberately deviates from the typed-deserialization rule (unknown-at-compile-time shape is the point of a JSON mock); sample-only, documented in the spec's exceptions section.
- Envelope detection on POCOs relies on reflection; under trimming it degrades safely to wrapping the object whole. Dictionary/Expando is the documented trim-safe path.

## PR Checklist

- [x] Tested code with current supported SDKs (net9.0; Playground on `net9.0-desktop`)
- [x] Docs added: `doc/Learn/Mvux/FeedView.md` — "Mock data and design-time sources"
- [x] Unit tests (`Given_MockFeed`, `Feed.Value` cases — full Reactive suite green: 1443 passed) and UI-host tests (`Given_FeedView`, run by runtime-test stages)
- [ ] Wasm UI Tests screenshots compare — to validate on CI
- [x] Contains **NO** breaking changes (additive only; previously-discarded sources now render)
- [ ] Release Notes — deferred until this graduates from POC
- [ ] Associated with an issue — **none yet (POC)**, see note at top

## Other information

Internal Issue (If applicable): none (POC)

### DIFF
--- a/doc/Learn/Mvux/FeedView.md
+++ b/doc/Learn/Mvux/FeedView.md
@@ -55,6 +55,50 @@ The `Source` property of the `FeedView` is data bound to the `CurrentContact` pr
 
 In the above example, [`Data`](#data) is a property of the `FeedViewState` instance that the `FeedView` creates from the `IFeed` and sets as the `DataContext` for the various templates.
 
+#### Mock data and design-time sources
+
+The `Source` property also accepts values which are not feeds, making it easy to preview a page with mock data without writing a Model. A plain object (a POCO, an anonymous object, or an `ExpandoObject`/dictionary created from JSON) is wrapped as a single-message feed whose value is the object, rendered by the [`ValueTemplate`](#valuetemplate-default-template):
+
+```xml
+<mvux:FeedView>
+    <mvux:FeedView.Source>
+        <models:Person Age="34" />
+    </mvux:FeedView.Source>
+    <DataTemplate>
+        <TextBlock Text="{Binding Data.Age}"/>
+    </DataTemplate>
+</mvux:FeedView>
+```
+
+To mock the other states of the `FeedView` (loading, error, empty), provide a mock *envelope*: an object whose members are limited to the following (matched case-insensitively, so camelCase JSON works):
+
+| Member | Aliases | Effect |
+| ------ | ------- | ------ |
+| `Data` | — | `null` renders the [`NoneTemplate`](#nonetemplate); any other value renders the `ValueTemplate` with that value. If the value is itself a feed, that feed is used directly. |
+| `Progress` | `IsProgress`, `InProgress` | `true` renders the [`ProgressTemplate`](#progresstemplate), mocking an in-flight load. |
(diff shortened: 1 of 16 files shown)


nickrandolph:
Closing this for now as we're not going in this direction

Issue #3155, opened 2026-08-24
MSAL authentication on every head: runtime dispatch, WebAssembly token cache, device runtime tests


kazo0:
Tracking issue for the stacked delivery of the MSAL authentication work that was opened as a single PR in #3139 and is being split so each layer reviews on its own. Parts that were independent already landed: #3151 (build hygiene), #3152 (auth core), #3153 (storage).

Problem, in one paragraph: `Uno.Extensions.Authentication.MSAL` is wrong in a different way on almost every head. Skia Android/iOS heads load the plain `netX.0` build of the provider (Uno.Sdk substitutes it), and that build was a stub, so `AddMsal` registers nothing. Redirect URIs are not derived, so apps hand-write `#if` blocks to sign in. On WebAssembly the token cache is memory-only and the app's redirect URI is silently overridden. And none of it had tests or a CI lane to run them on.

Related: #3025, #2438, #2346, unoplatform/uno#20601 (fixed by unoplatform/uno#24055).

Layers, bottom to top - each has its own sub-issue and PR; review bottom-up:

1. `ci:` device runtime-test lanes (desktop, WebAssembly, Android, iOS)
2. `feat(msal):` runtime platform dispatch, redirect defaults, storage defaults, sign-out, plus the unattended StubEntra test harness
3. `feat(msal):` WebAssembly token-cache persistence through `IKeyValueStorage`
4. `build:` require the Uno version containing the Skia-mobile asset-selection fix

The stack closes this issue when the top layer merges.

Issue #3025, opened 2026-02-07
MsalAuthenticationProvider does not work on MacOS out of the box


Flimzes:
## Current behavior
After creating a new Uno Cross Platform solution on MacOs using rider, adding MSAL authentication, and running the app, the app shows a login page
If we have added a ClientId and RedirectUri, the login button does nothing, and the console logs
fail: Uno.Extensions.Authentication.MSAL.MsalAuthenticationProvider[0]
      SetupStorage - Error setting up storage for MSAL - Value cannot be null. (Parameter 'keyChainServiceName')
Trying to add keyChainServiceName into appsettings under the msal config section does not work
Trying to find any arguments in AddMsal relating to this settings returns nothing.

## Expected behavior
The login button should work

## How to reproduce it (as minimally and precisely as possible)
Use Rider on Macos, install the uno plugin, run the uno command line tool to set up the environment correctly until it says everything OK!
Create a new Solution using the Uno Cross Platform template, and enable MSAL authentication
Add a clientId, a RedirectUri, and a KeychainSecurityGroup to the appsettings.json
Run the project
Attempt to log in using the login button

## Environment
Mac OS Tahoe 26.2
Rider 2025.3.2
Uno Plugin 1.9.12
.net 10

Nuget Package (s):
The default installed by the template

Package Version(s):
The default installed by the template

Affected platform(s):

- [ ] iOS
- [x] macOS (AppKit)
- [ ] Mac Catalyst
- [ ] Android
- [ ] WebAssembly
- [ ] Windows
- [ ] Skia (WPF)
- [ ] Skia (GTK on Linux/macOS/Windows)
- [ ] Skia (Linux Framebuffer)
- [ ] Build tasks

Visual Studio:

- [ ] 2019 (version: )
- [ ] 2022 (version: )
- [ ] Visual Studio Code (version: )
- [ ] Rider Windows (version: )
- [x] Rider macOS (version: ) 2025.3.2

Relevant plugins:

- [ ] Resharper (version: )

## Anything else we need to know?
Tried googling this error with 0 results, tried arguing with an LLM about it for 30 minutes with 0 results

Issue #2438, opened 2024-04-24
[Desktop] Could not load file or assembly Uno.Extensions.Authentication.MSAL.WinUI


eriklimakc:
### Current behavior

When adding `AuthenticationMsal` to `<UnoFeatures>` and launching on Desktop an exception is thrown:

```
System.IO.FileNotFoundException: 'Could not load file or assembly 'Uno.Extensions.Authentication.MSAL.WinUI, Version=255.255.255.255, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.'
```

### How to reproduce it (as minimally and precisely as possible)

Repro: [msal.zip](https://github.com/unoplatform/uno/files/15094365/msal.zip)
Binlog: [msal_Debug_AnyCPU_Build_2024-04-24T13_41_23.9873602+01_00.zip](https://github.com/unoplatform/uno/files/15094369/msal_Debug_AnyCPU_Build_2024-04-24T13_41_23.9873602%2B01_00.zip)

### Works on UWP/WinUI

Yes

### NuGet package version(s)

`Uno.Templates 5.2.28`

```json
{
  "msbuild-sdks": {
    "Uno.Sdk": "5.2.83"
  }
}
```

```xml
<UnoExtensionsVersion>4.1.14</UnoExtensionsVersion>
<UnoToolkitVersion>6.0.18</UnoToolkitVersion>
<UnoThemesVersion>5.0.13</UnoThemesVersion>
<UnoCSharpMarkupVersion>5.2.13</UnoCSharpMarkupVersion>
```

Issue #2346, opened 2024-06-13
Inplicit dependency on Microsoft.Identity.Client.Extensions.Msal breaks Wasm builds


jeromelaban:
## Current behavior

https://github.com/unoplatform/uno.extensions/commit/14979905a3a4c35a2f72a546d3471ffa32da6bb9 removed the Microsoft.Identity.Client.Extensions.Msal dependency from net8.0, but this dependency is still required to get Wasm apps to build.

This implicitly becomes a hidden dependendy, and while uno.sdk can be adjusted to work around that, apps not using uno.sdk will fail with this:

```
C:\Users\jela\.nuget\packages\uno.wasm.bootstrap\8.0.16\build\Uno.Wasm.Bootstrap.targets(213,3): [00:00:01.0838302] Unhandled exception. System.Exception: Could not resolve Microsoft.Identity.Client.Extensions.Msal
```

To fix this, we need a similar behavior to what's done with lottie, but when wasm is detected:

https://github.com/unoplatform/uno/blob/4533fe26120d4eeb7ce4198e08d5593c91d69f50/src/AddIns/Uno.UI.Lottie/buildTransitive/Uno.UI.Lottie.targets#L9-L29

PR #3161, opened 2026-08-24
feat(msal): platform-correct sign-in, storage defaults and sign-out, tested against a stub tenant

### LINKED ISSUE #3157: MSAL: platform-correct sign-in, storage defaults and sign-out, with an unattended test harness
Layer 2/4 of #3155. Runtime platform dispatch instead of #if (fixes the stub provider on Skia Android/iOS), derived redirect URIs with documented precedence, keychain/keyring storage defaults (#3025), sign-out of every account, refresh semantics, InteractiveBuilder, MSAL 4.87.0, and the StubEntra/StubWebUi harness that drives real MSAL end to end with no network. The device lanes from layer 1 run that namespace. Deliberately deferred upward: WebAssembly token-cache persistence (layer 3) and the Uno version floor (layer 4).

kazo0:
GitHub Issue (If applicable): closes #3157 (layer 2/4 of #3155; relates to #3025, #2438, #2346)

**Stack 2/4 — base: `msal-1-ci-lanes`. Review only this layer's diff.** Depends on layer 1 for the device lanes that run its tests. Deliberately deferred upward: WebAssembly token-cache persistence (layer 3) and the Uno version floor (layer 4).

## PR Type

- Bugfix
- Feature

- **Skia Android/iOS** — `AddMsal` registers nothing. Uno.Sdk substitutes the plain `netX.0` build of the provider on those heads, and that build was a stub (`UNO_EXT_MSAL` undefined), so every auth call fails with "No providers specified". The consumer-side `Package.targets` gated the define on repo-internal properties that are empty in apps, so apps got the define on every TFM including the stubbed ones.
- **Mac Catalyst** — `AddMsal` returns silently and fails the same way, much later.
- **Android, iOS, desktop** — no redirect URI is applied; apps hand-write an `#if ANDROID / #elif IOS / #else` block to sign in.
- **macOS and Linux desktop** — `MsalCacheHelper.CreateAsync` throws `ArgumentNullException` (#3025): `SetupStorage` supplies no keychain/keyring properties. Caught, logged, no fallback.
- **Everywhere** — a failed silent refresh is reported as success (empty access token); sign-out removes only the first account; `LoginAsync` re-wraps MSAL exceptions and loses the type; the per-request interactive modifiers (`WithPrompt`, `WithLoginHint`, `WithCustomWebUi`, …) are unreachable from the public surface. And none of it has a test.

## What is the new behavior?

**Platform selection is runtime dispatch** (`OperatingSystem.Is*`, with `IsMacCatalyst()` excluded from the iOS path), so the substituted plain-TFM assembly behaves correctly on Skia mobile. `UNO_EXT_MSAL` is defined for every platform that ships a functional provider and `Package.targets` applies the same allow-list in consuming projects. Mac Catalyst throws `PlatformNotSupportedException` at registration.

**Redirect URIs are derived per platform** — `msal{ClientId}://auth`, `msauth.{BundleId}://auth`, the `WebAuthenticationBroker` URI, `WithDefaultRedirectUri()` — with documented precedence: platform default < configuration < the app's `Builder(...)` callback, which now runs *after* `WithUnoHelpers()` so it wins everywhere (it used to be overwritten on WebAssembly). `UseDefaultPlatformRedirectUri = false` opts out.

**Token-cache persistence is platform-aware:** `MsalStorageDefaults` supplies the keychain/keyring properties (configurable), mobile lets MSAL persist natively, and an opt-in `AllowUnprotectedTokenCacheFallback` keeps sign-in state on a machine with no secure store. `Microsoft.Identity.Client` moves to 4.87.0 for macOS desktop under .NET 10.

**Lifecycle:** sign-out removes every account; a refresh that cannot renew the session returns no tokens (the service, from #3152, then signs out) while a transient failure keeps them; the 5-minute interactive timeout defaults only on desktop; MSAL exceptions are rethrown untouched. **Additive API:** `IMsalAuthenticationBuilder.InteractiveBuilder(...)`.

**Tests.** `Uno.Extensions.Authentication.MSAL.Tests` (unit, linked sources): redirect derivation and precedence, storage defaults. `Uno.Extensions.Authentication.MSAL.UI.Tests`: `StubEntra` fakes the tenant at the HTTP layer and `StubWebUi` the browser, so login → cache → silent refresh → rejected refresh → transient refresh → sign-out → cancellation all run through **real MSAL** (PKCE, code redemption, cache writes) with no network and no human. The device lanes now run this namespace. The vendored `Microsoft.Identity.Client.Extensions.Msal/*.wasm.cs` stubs were dead code and are removed; TestHarness drops its hard-coded redirect that the provider now honours.

### Behaviour changes (documented in `UpdatingExtensions.md`)

Catalyst `AddMsal` throws; the app's `Builder(...)` wins over the platform default and `WithUnoHelpers()`; unrenewable refresh signs out; sign-out removes all accounts; exceptions untouched; desktop-only timeout default.

## PR Checklist

- [x] Docs have been added/updated which fit documentation template
- [x] Unit Tests and/or UI Tests for the changes have been added
- [ ] Wasm UI Tests are not showing any unexpected differences
- [ ] Contains **NO** breaking changes — see the behaviour changes above
- [ ] Updated the Release Notes

## Other information

Verified locally: MSAL.Tests 27/27; package surface builds warning-free on all seven TFMs; desktop runtime suite 13 MSAL + 8 storage = 21/21. Every bug fix has a test that was red before it (recorded in `specs/012` and `specs/010` — 012 because `main` took 009 for the hot-reload route refresh while this stack was open). Still open in spec 010: live on-device validation of the Skia-mobile fix — the runtime-test heads use `ProjectReference`, so the SDK's substitution never fires in CI.

https://claude.ai/code/session_01EqfQh5ovrPEvFGcERLiLhH


src/Uno.Extensions.Authentication.MSAL.UI.Tests/Given_MsalAuthentication.cs
```diff
@@ -0,0 +1,408 @@
...
+						// removes a network-shaped failure mode from mobile CI.
+						.WithInstanceDiscovery(false)
+						.WithHttpClientFactory(tenant.HttpClientFactory))
+					// Deliberately no WithRedirectUri: the provider's platform default applies, which
+					// is the behaviour under test. Hard-coding http://localhost was desktop-shaped and
+					// is not a valid redirect on iOS, where MSAL expects the msauth scheme - a likely
+					// reason all 7 tests failed there once the harness finally ran (build 227612).
+					// StubWebUi echoes back whatever redirect URI MSAL hands it, so any value works.
+					.InteractiveBuilder(interactive => interactive
+						.WithCustomWebUi(webUi))))
```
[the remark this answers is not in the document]
kazo0: False positive. Line 124 calls `AddMsal(window, ...)` — the `Window` overload, which is the non-obsolete one. The `[Obsolete]` attribute is on the parameterless-window overload two declarations above; the analyzer resolved the wrong overload. The build is green under `TreatWarningsAsErrors`, which would have turned a real CS0618 into an error. No code change.

src/Uno.Extensions.Authentication.MSAL/MsalAuthenticationBuilderExtensions.cs
```diff
@@ -37,6 +37,47 @@ Action<PublicClientApplicationBuilder> build
...
+	/// <returns>
+	/// The <see cref="IMsalAuthenticationBuilder"/> that was passed in.
+	/// </returns>
+	public static IMsalAuthenticationBuilder InteractiveBuilder(
+		this IMsalAuthenticationBuilder builder,
+		Action<AcquireTokenInteractiveParameterBuilder> build
+		)
+	{
+#if !UNO_EXT_MSAL
+		return builder;
```
ajpinedam: In this file src/Uno.Extensions.Authentication.MSAL/HostBuilderExtensions.cs it changed to throw an exception when the feature is not supported, should we do the same here?

rajamatt (approved):

src/Uno.Extensions.Authentication.MSAL/MsalStorageDefaults.cs
```diff
@@ -0,0 +1,68 @@
...
+	/// on macOS (keychain) and Linux (keyring). Without these the cache helper throws
+	/// <see cref="ArgumentNullException"/> on those platforms (https://github.com/unoplatform/uno.extensions/issues/3025).
+	/// </summary>
+	internal static void ApplyDefaults(
+		StorageCreationPropertiesBuilder builder,
+		string? clientId,
+		string? configuredServiceName,
+		string? configuredAccountName,
+		bool isMacOS,
+		bool isLinux)
```
ajpinedam: nit: would not an Enum fit better here considering these two flags could be both ON (by mistake) and it would default to macOS.

ajpinedam (approved):

src/Uno.Extensions.Authentication.MSAL/MsalAuthenticationBuilderExtensions.cs
kazo0: Good question — the answer turned out to be "neither". That `#if !UNO_EXT_MSAL` build exists only for Mac Catalyst, and there `AddMsal` already throws `PlatformNotSupportedException` before the `configure` callback can run, so `Builder`, `InteractiveBuilder`, `Storage` and `Scopes` could never be reached on the stub build: the `return builder;` branches were dead code, and making them throw would have been more dead code. Their real bodies compile on every TFM (`MsalAuthenticationSettings` is not gated), so 3243807af deletes the four `#if` blocks instead. Verified by building the provider for `net9.0-maccatalyst` and `net9.0` in Release. `WithWebRedirectUri` keeps its stub: it hangs off `PublicClientApplicationBuilder`, so it *is* reachable without `AddMsal`.

src/Uno.Extensions.Authentication.MSAL/MsalStorageDefaults.cs
kazo0: Agreed — done in 3243807af. `ApplyDefaults` now takes `MsalSecureStore { None, MacKeychain, LinuxKeyring }` and switches on it; `MsalStorageDefaults.ForCurrentOS()` is the single place that maps `OperatingSystem.IsMacOS()/IsLinux()` to a value, so the both-true state cannot exist and the provider call site no longer carries the two probes. Tests updated (27/27 in this layer, 39/39 at the top of the stack).

Issue #3174, opened 2026-09-04
Unsubscribing from a BindableListFeed during finalization kills the process when hosted in a collectible AssemblyLoadContext


nickrandolph:
## Summary

Removing an event handler from a `BindableListFeed<T>` / `BindableCollection` resolves the current dispatcher, which reads a static `ThreadLocal<IDispatcher>`. If that `ThreadLocal` has been disposed, the read throws `ObjectDisposedException`.

Because the removal is reached **from a finalizer**, and an exception escaping a finalizer is unrecoverable, this **terminates the process**.

It is reproducible whenever `Uno.Extensions.Reactive.UI` is loaded into a collectible `AssemblyLoadContext` that is later unloaded — a plugin/preview host, a designer surface, or anything that hot-loads user assemblies.

```
[UNHANDLED_DOMAIN_EXCEPTION] IsTerminating=True ObjectDisposedException: Cannot access a disposed object.
Object name: 'System.Threading.ThreadLocal`1[[Uno.Extensions.IDispatcher, Uno.Extensions.Core, Version=255.255.255.255, Culture=neutral, PublicKeyToken=null]]'.
   at System.Threading.ThreadLocal`1.GetValueSlow()
   at Uno.Extensions.Reactive.Bindings.BindableListFeed`1.remove_CollectionChanged(NotifyCollectionChangedEventHandler value)
   at Microsoft.UI.Xaml.Controls.ItemsSourceView.Finalize()
   at System.GC.RunFinalizers()
Unhandled exception. System.ObjectDisposedException: Cannot access a disposed object.
...
```

## Mechanism

Verified by reading the source on `servicing/7.4`, not inferred from the stack trace:

1. `DispatcherQueueProvider._value` is a `private static readonly ThreadLocal<IDispatcher?>`
   — `src/Uno.Extensions.Reactive.UI/Utils/Dispatching/DispatcherQueueProvider.cs:16`
2. `ModuleInitializer` wires it in as the process-wide resolver
   — `src/Uno.Extensions.Reactive.UI/ModuleInitializer.cs:33`:
   `DispatcherHelper.GetForCurrentThread = DispatcherQueueProvider.GetForCurrentThread;`
3. Unsubscribing walks from the feed down to that resolver:
   - `BindableListFeed<T>.CollectionChanged` `remove => _items.CollectionChanged -= value;`
     — `Presentation/Bindings/BindableListFeed.CollectionView.cs:138`
[...]

PR #3175, opened 2026-09-04
fix(reactive): treat a disposed dispatcher provider as no dispatcher

nickrandolph:
Fixes #3174 — read that issue for the full mechanism; this is the fix and its evidence.

Draft because the reproduction that motivated it is an end-to-end collectible-`AssemblyLoadContext` host that lives downstream, so CI here can only prove the unit-level behaviour. See **Verification** for exactly what is and is not established.

## The bug in one paragraph

Removing an event handler from a `BindableCollection` resolves the current dispatcher so it can find that thread's `DataLayer` (`BindableCollection.cs:239` → `DispatcherLocal.Value` → `_schedulersProvider()`), and that resolution reads `DispatcherQueueProvider`'s static `ThreadLocal<IDispatcher>`. `ThreadLocal<T>` disposes itself from its own finalizer, so once the assembly owning that static is unloaded with a collectible ALC, the read throws `ObjectDisposedException`. Removal is reached **from** a finalizer — `~ItemsSourceView()` disposes its collection-changed listener, which unsubscribes — and an exception escaping a finalizer is unrecoverable: it terminates the process.

## The change

**`DispatcherLocal` (core)** — every current-dispatcher resolution now goes through `FindCurrentDispatcher()`, which reports `null` on `ObjectDisposedException`. All four call sites are covered: `Value` get, `Value` set, `GetValue`, `TryGetValue`. (`TryGetValue` matters even though the owner is passed explicitly — it still reads the *current* dispatcher to decide whether creation from another thread is permitted.)

Null is the truthful answer here, not a degraded one: a finalizer thread is not a UI thread, so there is genuinely no dispatcher to find. `DispatcherLocal` already has the background-value path for exactly that case, and `allowBackgroundValue` is **never overridden anywhere in this repo** (default `true`, `DispatcherLocal.cs:41`; the only production construction is `BindableCollection.cs:132`, which passes only `allowCreationFromAnotherThread: true`) — so that path is always available. Had it been `false`, this fix would merely have swapped a fatal `ObjectDisposedException` for a fatal `InvalidOperationException`, which is why it is called out rather than assumed.

**`DispatcherQueueProvider` (UI)** — `GetForCurrentThread()` guards the same way, so the provider cannot throw even for a consumer that resolves it directly rather than through `DispatcherLocal`. Defence in depth; the core fix is the load-bearing one.

**Only `ObjectDisposedException`** is treated this way. Any other resolution failure stays fatal — a misconfigured provider must not silently degrade every consumer to background values. There is a test for that specifically.

Chosen this way because `DispatcherLocal` is the layer that *depends* on "resolve the current dispatcher", it is reachable from every one of the eight removal paths in `BindableCollection` (lines 239, 245, 335, 338, 344, 347, 353, 356 — `VectorChanged` and the selection handlers resolve on removal too, so `CollectionChanged` is not the only exposure), and it takes an injectable provider, which makes the fix testable without disposing a process-wide static from a test.

## Deliberately not done here

Removal should arguably not resolve a dispatcher **at all**. `DispatcherLocal.ForEachValue` already iterates existing layers without resolving or creating anything, so the removal accessors could apply across all layers and be naturally idempotent — which would additionally fix removing a handler from a different thread than the one that added it, and would avoid constructing a `DataLayer` on a finalizer thread.

That is a behavioural change to event-removal semantics across eight accessors, including token-based overloads whose tolerance for unknown tokens I have not audited. It does not belong on a servicing branch. Happy to open it against `main` if you want it.

## Verification

**Red/green, run both ways:**

| Run | Result |
|---|---|
| 3 new tests with the fix | **6 passed / 0 failed** (`Given_DispatcherLocal`) |
| Same tests with `DispatcherLocal.cs` reverted, tests kept | **2 failed / 4 passed** — `When_ProviderDisposed_Then_ValueFallsBackToBackground`, `When_ProviderDisposed_Then_TryGetValueDoesNotThrow` |
| Full `Uno.Extensions.Reactive.Tests` | **1422 passed / 0 failed / 19 skipped** |
| `dotnet build` of the test project | 0 errors |

The third new test — `When_ProviderThrowsOtherError_Then_ItPropagates` — passes in *both* directions by design: it asserts the exception types that must **keep** throwing, so it guards the narrowness of the `catch` rather than the fix itself.

**Not established here, and worth being plain about:** no end-to-end proof that the process no longer dies. That needs a host loading `Uno.Extensions.Reactive.UI` into a collectible ALC, unloading it, and forcing finalizers — the setup described in the issue. The unit tests reproduce the *provider state* that causes it (a provider that throws `ObjectDisposedException`) and prove the resolution path now tolerates it; they do not reproduce the ALC teardown itself. I will confirm end-to-end against a downstream host once this is in a 7.4-dev package and report back on the issue.

One compile detail, in case it comes up in review: the XML doc on `FindCurrentDispatcher` refers to `DispatcherQueueProvider` in prose rather than with a `cref`, because it lives in the UI assembly and core does not reference it — a `cref` fails the build with CS1574.

### DIFF
--- a/src/Uno.Extensions.Reactive.Tests/Utils/Dispatching/Given_DispatcherLocal.cs
+++ b/src/Uno.Extensions.Reactive.Tests/Utils/Dispatching/Given_DispatcherLocal.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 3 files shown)


iurycarlos:
Confirming this from the consuming side: we hit exactly the crash in #3174 — a host that loads apps into collectible `AssemblyLoadContext`s and forces a finalizer wait after each unload — and this PR addresses it.

Fixing it centrally in the dispatcher lookup rather than at the individual call sites looks like the right call to us. Letting anything other than `ObjectDisposedException` propagate is a good detail; we had missed that in our own attempt and would have degraded a misconfigured provider to background values silently.

**One observation, offered as optional follow-up rather than a change request.**

After this fix, a removal reached from a finalizer resolves to "no dispatcher" and falls through to the background data layer — which `When_ProviderDisposed_Then_ValueFallsBackToBackground` shows being *created* on demand. So the factory (`DataLayer.Create(...)`) now runs on the finalizer thread during context teardown.

That is a much smaller problem than a process abort, but it is not free: it executes construction logic at finalization time, and if anything inside that factory reaches another already-finalized static, the same fatal shape returns from a different line. It also means unsubscribing still materialises state, which reads oddly given the "why removal resolves a dispatcher at all" section of #3174.

We have a branch that closes that gap, if it is of interest:

- a non-creating `DispatcherLocal<T>.TryGetCurrentValue` that returns the current thread's value only when it already exists and never invokes the factory;
- the eight removal paths in `BindableCollection` (collection-changed, property-changed, vector-changed and the four selection ones) routed through it, so removal from a layer that was never materialised is simply a no-op.

Verified against `main`: 1435 unit tests pass, Release builds of `Uno.Extensions.Reactive` and `Uno.Extensions.Reactive.WinUI` clean. Happy to open it as a follow-up on top of this PR if you want it, or to drop the idea entirely — it is hardening, not a fix, and this PR is what stops the crash.

We had opened #3177 / #3178 before spotting #3174 and this PR; both are now closed as duplicates. Apologies for the noise.

Jen-Uno (approved):

nickrandolph:
@mergify backport main

PR #3180, opened 2026-09-10
fix: ship the mock generator in the package and null-inject every constructor parameter

rajamatt:
GitHub Issue (If applicable): related to https://github.com/unoplatform/studio.live/issues/4074

## PR Type

- Bugfix
- Build or CI related changes

`Uno.HotTesting.Reactive` (spec 013) is published with its runtime only: `Uno.HotTesting.Reactive.Generator` is never packed, so referencing the package generates no `{Model}Mock` / `{Vm}Mock` types.

The generator also emits `new {Vm}(default!)` with exactly one argument. A model whose constructor takes more than one dependency (a service plus a navigator or a messenger, the shape MVUX messaging produces) gets a `Create` factory that does not compile.

## What is the new behavior?

- The package ships the generator under `analyzers/dotnet/cs`, the way `Uno.Extensions.Reactive` ships the MVUX generator: a `ReferenceOutputAssembly="false"` project reference for build ordering and the generator DLL packed as a `None` item. The generator project joins the package-only and Reactive solution filters so package builds include it.
- `Create` targets the public view-model constructor with the fewest parameters (the protected model-wrapping constructor is never a candidate) and passes `default!` for each parameter. Output is unchanged for single-parameter models.
- Fixture `MenuModel` (two constructor parameters) in the two-project mocking app, test `When_ModelHasSeveralCtorParameters_Then_CreateNullInjectsEachOne`, one sentence in `doc/Reference/Reactive/testing.md`, and a spec 013 history note.

## Other information

Verified locally: a Release pack of `Uno.HotTesting.Reactive` contains `analyzers/dotnet/cs/Uno.HotTesting.Reactive.Generator.dll` next to the runtime; the `Uno.Extensions.Reactive.Tests` mocking tests and `Uno.HotTesting.Reactive.Tests` pass.

### DIFF
--- a/src/Uno.Extensions.Reactive.Tests.MockingApp/MenuModel.cs
+++ b/src/Uno.Extensions.Reactive.Tests.MockingApp/MenuModel.cs
@@ -1,0 +1,44 @@
+using System.Collections.Immutable;
+using System.Threading;
+using System.Threading.Tasks;
+using Uno.Extensions.Reactive;
+
+namespace Uno.Extensions.Reactive.Tests.MockingApp;
+
+public interface IMenuService
+{
+	Task<IImmutableList<string>> GetItems(CancellationToken ct);
+}
+
+public interface IMenuNavigator
+{
+	ValueTask GoBack(CancellationToken ct);
+}
+
+/// <summary>
+/// Fixture for a model whose constructor takes more than one dependency: the generated Create must
+/// null-inject each parameter. Also carries a derived feed and an independent state for later coverage.
+/// </summary>
+public partial class MenuModel
+{
+	private readonly IMenuService _service;
+	private readonly IMenuNavigator _navigator;
+
+	public MenuModel(IMenuService service, IMenuNavigator navigator)
+	{
+		_service = service;
+		_navigator = navigator;
+	}
+
+	// service-dependent input (list)
+	public IListFeed<string> Items => ListFeed.Async(async ct => await _service.GetItems(ct));
+
+	// derived over the service-dependent list
+	public IFeed<int> ItemsCount => Items.AsFeed().Select(items => items.Count);
+
+	// independent state — not part of the mock
+	public IState<string> Filter => State<string>.Value(this, () => string.Empty);
+
+	// command → IAsyncCommand GoBack on the VM; the navigator is only reached when it executes
+	public ValueTask GoBack(CancellationToken ct) => _navigator.GoBack(ct);
+}
--- a/src/Uno.Extensions.Reactive.Tests.MockingApp/MenuModel.cs
+++ b/src/Uno.Extensions.Reactive.Tests.MockingApp/MenuModel.cs
@@ -15,9 +15,17 @@ public interface IMenuNavigator
 	ValueTask GoBack(CancellationToken ct);
 }
 
+public interface IMenuSnapshot
+{
+	IImmutableList<string> Items { get; }
+}
+
 /// <summary>
-/// Fixture for a model whose constructor takes more than one dependency: the generated Create must
-/// null-inject each parameter. Also carries a derived feed and an independent state for later coverage.
+/// Fixture for the constructor shapes the generated Create must handle: more than one dependency (each
+/// parameter null-injected) and two public constructors of equal arity (the parameter type must be spelled
+/// out, otherwise <c>new MenuViewModel(default!, default!)</c> is ambiguous, CS0121). The derived feed and
+/// the independent state pin the record's member classification: the derived feed is an optional override,
+/// the independent state is left out.
 /// </summary>
 public partial class MenuModel
 {
@@ -28,6 +36,11 @@ public partial class MenuModel
 	{
 		_service = service;
 		_navigator = navigator;
+	}
+
+	public MenuModel(IMenuSnapshot snapshot, IMenuNavigator navigator)
+		: this(new SnapshotMenuService(snapshot), navigator)
+	{
 	}
 
 	// service-dependent input (list)
@@ -41,4 +54,16 @@ public partial class MenuModel
 
 	// command → IAsyncCommand GoBack on the VM; the navigator is only reached when it executes
 	public ValueTask GoBack(CancellationToken ct) => _navigator.GoBack(ct);
+}
+
+internal sealed class SnapshotMenuService : IMenuService
+{
+	private readonly IMenuSnapshot _snapshot;
+
+	public SnapshotMenuService(IMenuSnapshot snapshot)
+	{
+		_snapshot = snapshot;
+	}
+
+	public Task<IImmutableList<string>> GetItems(CancellationToken ct) => Task.FromResult(_snapshot.Items);
 }
--- a/src/Uno.Extensions.Reactive.Tests/Mocking/Given_GeneratedMock.cs
+++ b/src/Uno.Extensions.Reactive.Tests/Mocking/Given_GeneratedMock.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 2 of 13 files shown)


src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -146,6 +146,18 @@ private sealed class FeedMember
...
+		var ctor = vm.Constructors
+			.Where(c => !c.IsStatic && c.DeclaredAccessibility == Accessibility.Public)
+			.OrderBy(c => c.Parameters.Length)
+			.FirstOrDefault();
+		if (ctor is null)
+		{
+			return null;
+		}
+
+		var ctorArguments = string.Join(", ", ctor.Parameters.Select(p => $"default! /* {p.Name} */"));
```
dr1rrb: `default!` is untyped, so the emitted call cannot resolve between same-arity overloads.

The MVUX generator emits **one public view-model constructor per public model constructor** (`ViewModelGenTool_3`, the `model.Constructors` projection). A model with two single-parameter constructors — say `MenuModel(IMenuService)` and `MenuModel(IMenuNavigator)` — therefore produces a view-model with two single-parameter public constructors, and the generated

```csharp
var vm = new MenuViewModel(default!);
```

is ambiguous: `default` is convertible to every type, so overload resolution has nothing to discriminate on and the consumer project fails with `CS0121`. The arity is right, the target is not.

Emitting the parameter type makes the call unambiguous and costs nothing:

```csharp
var ctorArguments = string.Join(", ", ctor.Parameters.Select(p =>
    $"default({p.Type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat)})!"));
```

The `/* name */` comment is a nice touch for readability — worth keeping alongside the cast.

I could not reproduce this on the fixture app (it has a single constructor per model), so treat it as a language-level risk rather than an observed failure — but it is the exact shape this PR is fixing, one step further.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -146,6 +146,18 @@ private sealed class FeedMember
 			return null;
 		}
 
+		// Create null-injects the public view-model constructor with the fewest parameters. The generated VM
+		// mirrors the model's constructors; its protected model-wrapping constructor is never a candidate.
+		var ctor = vm.Constructors
+			.Where(c => !c.IsStatic && c.DeclaredAccessibility == Accessibility.Public)
+			.OrderBy(c => c.Parameters.Length)
+			.FirstOrDefault();
+		if (ctor is null)
```
dr1rrb, line 168: Silent bail-out: a model with mockable inputs but no public view-model constructor now generates **nothing**, with no diagnostic.

The consumer just observes that `{Vm}Mock` does not exist, with no way to tell whether the model was skipped, the opt-out was picked up, or the generator failed. The earlier `return null` above is benign (there is genuinely nothing to mock); this one hides a real, actionable condition.

Spec 013 already reserved `MOCK0001` for exactly this class of situation (`specs/013-mvux-mocking-previews/implementation.md`, diagnostics section). Reporting a diagnostic here — and still returning — would turn a puzzling absence into a fixable message.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -146,6 +146,18 @@ private sealed class FeedMember
 			return null;
 		}
 
+		// Create null-injects the public view-model constructor with the fewest parameters. The generated VM
+		// mirrors the model's constructors; its protected model-wrapping constructor is never a candidate.
+		var ctor = vm.Constructors
+			.Where(c => !c.IsStatic && c.DeclaredAccessibility == Accessibility.Public)
+			.OrderBy(c => c.Parameters.Length)
+			.FirstOrDefault();
```
dr1rrb, line 167: Tie-break is unspecified when several public constructors share the fewest parameters.

`OrderBy` is stable in LINQ-to-objects, so the winner is whichever comes first in `vm.Constructors` — and Roslyn does not guarantee a stable order for `INamedTypeSymbol.Constructors` across compilations or versions. For a source generator that means the emitted code can change without the source changing.

With the typed-`default` fix from the other comment the choice becomes observable (different parameter types in the output), so it is worth making deterministic — e.g. `.ThenBy(c => string.Join(",", c.Parameters.Select(p => p.Type.ToDisplayString())), StringComparer.Ordinal)`.

src/Uno.Extensions.Reactive.Tests/Mocking/Given_GeneratedMock.cs
```diff
@@ -81,4 +81,16 @@ public async Task When_CatalogEntry_Then_PinnedState()
 		var items = await CurrentItems(SourceContext.GetOrCreate(vm.Model), vm.Model.Steps);
 		items.Should().BeEquivalentTo(new[] { 1, 2, 3 });
 	}
+
+	[TestMethod]
+	public async Task When_ModelHasSeveralCtorParameters_Then_CreateNullInjectsEachOne()
+	{
+		// MenuModel takes a service and a navigator: Create passes default for both, so it compiles and the
+		// mocked input flows without either dependency.
```
dr1rrb: The test name promises more than the assertions check.

`When_ModelHasSeveralCtorParameters_Then_CreateNullInjectsEachOne` asserts that the mocked input flows; nothing observes that *each* parameter was null-injected. The real proof of this fix is that the generated `Create` **compiles** at all — before the change, a two-parameter model made the whole test project fail to build.

That is a legitimate and valuable test, it is just a compile-time one. Either rename it to what it verifies (`..._Then_CreateCompilesAndInputFlows`) or add the missing observation, so the name stays a reliable index of coverage.

src/Uno.Extensions.Reactive.Tests.MockingApp/MenuModel.cs
```diff
@@ -0,0 +1,44 @@
...
+	private readonly IMenuNavigator _navigator;
+
+	public MenuModel(IMenuService service, IMenuNavigator navigator)
+	{
+		_service = service;
+		_navigator = navigator;
+	}
+
+	// service-dependent input (list)
+	public IListFeed<string> Items => ListFeed.Async(async ct => await _service.GetItems(ct));
```
dr1rrb, line 47: `ItemsCount` and `Filter` are introduced for coverage that this PR does not add (`for later coverage`, per the class comment).

They are not entirely idle — they do exercise the generator's member classification (derived, and independent state excluded from the mock), which is worth having in a fixture. But nothing asserts on them, so the comment reads as an open TODO in a shared fixture.

Either drop the two members until the tests that need them land, or state in the comment what they pin down today (classification coverage) so the next reader does not treat them as dead weight.

dr1rrb (commented):
Reviewed at `f161469ecf71f5fdadcee6399290fb568e18f0b1`.

Two good fixes: the analyzer was genuinely missing from the package, and the single `default!` was a real blocker for any model with more than one dependency. The packaging change also matches the repo convention exactly — `Uno.Extensions.Reactive`, `.Navigation` and `.Core` all pack their generator through the same `<None Include=... PackagePath="analyzers/dotnet/cs">`, so there is nothing to reconcile there (`ToolOfPackage` only drives the local NuGet-override copy, not the `.nupkg`).

Nothing blocking. The one comment worth acting on before merge is the untyped `default!`: it fixes the arity but not the overload resolution, so the same class of model — several public constructors of equal arity — still fails to compile. The rest are robustness and clarity notes.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
rajamatt: Done, now reports the reserved `MOCK0001`

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
rajamatt: Ties are broken now

src/Uno.Extensions.Reactive.Tests.MockingApp/MenuModel.cs
rajamatt: Kept them since they are not useless, added comment for clarity

kazo0 (approved):

PR #3184, opened 2026-09-15
fix(mocking): generate mocks when the models share the generator's compilation

rajamatt:

## PR Type

- Bugfix
- Feature
- Documentation content changes

The mocking generator shipped in `Uno.HotTesting.Reactive` (spec 013) reads the `[FeedDependency]` /
`[Model]` metadata the MVUX generator emits on a model. That works when the app is a **compiled
reference** — a test or preview project referencing it.

It does not work when the models live in the **same compilation** as the generator, which is what an app
referencing the package directly has. There the attributes are produced by a sibling source generator, and
a Roslyn generator cannot observe another generator's output, so no model was ever found. Referencing the
package in an app generated nothing at all, and reported nothing: the build succeeded and the types simply
were not there. Since the templates create a single project, this is the shape most consumers have.

Separately, the classifier recognized a service only through a constructor **parameter symbol** or a field
assigned in a **constructor body**. A positional record — `public partial record RecipeModel(IRecipeService
Service)`, the form the reference documentation itself uses — surfaces the parameter as a synthesized
property, so nothing matched, every feed was classified independent, and the mock came out with no required
members.

## What is the new behavior?

The generator now reaches models two ways. Compiled references are read from metadata exactly as before.
For models declared in the compilation being generated, it runs the same analysis over the source instead;
the view-model is not a symbol there either, so it is named and constructed from the model, which is what
the MVUX generator derives it from. Declared attributes still take precedence, so the hand-declaration
escape hatch is unaffected, and a project that can already see mocks generated in a reference does not emit
a second copy.

To keep the two generators from disagreeing about what a model's inputs are, the classification moved into
one file compiled into both. A primary constructor's parameters now map to the same-named member, which is
what fixes the positional-record case.

Also in this PR, from a review pass over the above:

- mock members are typed by the **feed interface** rather than the member's own type — `FeedMock.Empty`
  returns an `IFeed`, so a member typed as the state itself could not accept it (CS0266), which made every
  service-dependent `IState` / `IListState` uncompilable;
- emitted types carry the model's accessibility, so an `internal` model no longer produces a `public` mock
  (CS0050/CS0051);
- nested models are left to the metadata path, since their view-model is generated inside the containing
  partial and cannot be named from the namespace;
- `[assembly: EnableFeedMocking(IsEnabled = false)]` is honoured on both paths — it previously silenced the
  metadata but not the mocks;
- **MOCK0002** reports a model considered but carrying no mockable input, so the silent no-op this PR set
  out to remove does not survive in another form;
- constructor selection honours `[ReactiveBindable(false)]`, and may pick an internal constructor on the
  source path where the mirrored one is reachable;
- the generated hint name is sanitized and the user-supplied `ImplicitBindables` regex is bounded — either
  could previously throw out of `Execute` and drop every mock in the compilation;
- bound syntax trees are shared across models, and the walk skips assemblies that do not reference
  `Uno.Extensions.Reactive`, so an app build no longer realizes the whole metadata closure on every
  compilation;
- two solution entries for projects that no longer exist on disk are removed.

### Breaking change

A model declared as a positional record previously produced a mock with **no** required members, or no mock
at all. Its feeds are now recognized as service-dependent, so the mock gains `required` members and existing
`new {Model}Mock { }` / `Create()` call sites must supply them. Noted in
`doc/Reference/Reactive/testing.md`.

### Verification

`Uno.HotTesting.Reactive.Tests` 30/30 and the `Mocking` suite of `Uno.Extensions.Reactive.Tests` 19/19,
including new generator-driver cases for the source path, the opt-out, internal accessibility, a
service-dependent state, and MOCK0002. A new fixture project puts the models and the generator in one
compilation; a positional-record model was added to the existing two-project fixture, which had none.

Also exercised outside the repo against an app created from the templates: the mock is generated, the app
compiles, and the mocked state renders in a preview.

Not verified here: that the analyzer behaves identically once consumed from the packed NuGet rather than as
a local build.

## PR Checklist

- [x] Tested code with current [supported SDKs](../README.md#supported)
- [x] Docs have been added/updated which fit [documentation template](https://github.com/unoplatform/uno/blob/master/doc/.feature-template.md). (for bug fixes / features) 
- [ ] Contains **NO** breaking changes — see the breaking change noted above


src/Uno.Extensions.Reactive.Generator/Bindables/ViewModelGenTool_3.Mocking.cs
```diff
@@ -3,30 +3,35 @@
...
+/// When the opt-out is present, <see cref="GenerateMockingMetadata"/> returns an empty string, so the
 /// generated output is byte-identical (G5).
 /// </summary>
 internal partial class ViewModelGenTool_3
 {
-	private enum FeedKind { ServiceDependent, Derived, Independent }
+	// One per run, not per model: the analysis caches bound syntax trees, which the models share.
+	private FeedMockingAnalysis? _mockingAnalysis;
+
+	private FeedMockingAnalysis MockingAnalysis
```
dr1rrb: The generator should remain stateless

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -83,12 +200,128 @@ IEnumerable<INamedTypeSymbol> Walk(INamespaceOrTypeSymbol ns)
...
+
+			if (ReadReactiveBindable(type) is { } explicitlyEnabled)
+			{
+				// When the attribute is set the `partial` is not checked: the build must fail if it is missing.
+				return explicitlyEnabled;
+			}
+
+			return type.IsPartial()
+				&& implicitEnabled
+				&& patterns.Any(pattern => IsMatch(pattern, type.ToString()));
```
dr1rrb: ### ➡️ Medium
The model-discovery rule is re-implemented here rather than shared, while only the feed classification moved into `FeedMockingAnalysis`. This predicate mirrors `ViewModelGenTool_3.IsSupported`, and the two have already drifted: the patterns are parsed by hand out of `ConstructorArguments` here but read from the typed `ImplicitBindablesAttribute.Patterns` there, and the match is bounded here and unbounded there. The stated invariant, "compiled into both so the two cannot drift apart", therefore covers the rule whose divergence is silent, not the ones that make a model appear or disappear. Moving this predicate next to `FeedMockingAnalysis` in the shared `Mocking/` folder would extend the guarantee you already built the mechanism for.

Two smaller points in the same expression:

1. `type.ToString()` is evaluated inside the `patterns.Any` lambda, so the display string is rebuilt once per pattern for every top-level partial type of the compilation. Hoisting it into a local before the `Any` keeps one string per type.
2. The namespace recursion below is a second copy of the `Walk` in `EnumerateAttributedModels`; both sites are written in this diff, so one shared traversal is still cheap to do.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -127,105 +388,189 @@ private sealed class FeedMember
...
+	private static ModelMock? DescribeFromSource(INamedTypeSymbol model, FeedMockingAnalysis analysis)
+	{
+		var feedMembers = analysis.GetFeedMembers(model);
+		if (feedMembers.Count == 0)
+		{
+			return null;
+		}
+
+		var modelName = model.Name.TrimEnd("Model", StringComparison.Ordinal);
+		var viewModelName = $"{modelName}{ViewModelSuffix}";
```
dr1rrb: ### ➡️ Medium
The view-model name is re-derived locally here, a copy of `ViewModelGenTool_3.cs:86` (`TrimEnd("Model", Ordinal)` plus the `ViewModel` suffix, the two even keeping separate constants, `ViewModelSufix` and `ViewModelSuffix`). The emitted code then references `global::{ns}.{Name}ViewModel` by name. If the MVUX naming rule ever moves, the mock emits a reference to a type that does not exist, and the resulting compile error lands in generated code the consumer cannot edit. This is the rule whose divergence breaks the build, so it is the one most worth sharing. Deriving the name from the same helper both generators compile would close it.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -83,12 +200,128 @@ IEnumerable<INamedTypeSymbol> Walk(INamespaceOrTypeSymbol ns)
...
+			.FirstOrDefault(a => a.AttributeClass?.Name == ReactiveBindableAttribute);
+		if (attribute is null)
+		{
+			return null;
+		}
+
+		return attribute.ConstructorArguments.FirstOrDefault().Value as bool?
+			?? attribute.NamedArguments.FirstOrDefault(na => na.Key == "IsEnabled").Value.Value as bool?
+			?? true;
+	}
```
dr1rrb: ### ➡️ Medium
`ReactiveBindable`, `ImplicitBindables` and `EnableFeedMocking` are matched on `AttributeClass?.Name`, the simple class name, while `FeedDependency` and `Model` are resolved through `GetTypeByMetadataName` and compared with `SymbolEqualityComparer` at the top of `Execute`. A same-named attribute from any other namespace therefore silently drives model discovery, the constructor choice and the opt-out. Resolving these three the same way as the other two removes the ambiguity and keeps both generators agreeing on what counts as opted in.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -48,16 +78,103 @@ public void Execute(GeneratorExecutionContext context)
...
 		}
 	}
 
-	private static IEnumerable<INamedTypeSymbol> EnumerateModels(Compilation compilation, INamedTypeSymbol feedDep)
+	/// <summary>
+	/// The mock type already exists — generated in a referenced assembly that also has this package, or
+	/// hand-written. Emitting a second one would collide (CS0101) or be ambiguous at the use site.
+	/// </summary>
+	private static bool IsAlreadyDeclared(Compilation compilation, INamedTypeSymbol model)
+		=> compilation.GetTypeByMetadataName(MockMetadataName(model)) is not null;
```
dr1rrb: ### ➡️ Medium
`GetTypeByMetadataName` returns `null` when the name is declared in more than one referenced assembly, so this guard fails open in exactly the case it exists to cover: two references already declaring the mock, where a third copy is then emitted.

Related, on the name it builds: `MockMetadataName` drops the containing type, so two nested models sharing a simple name in one namespace both resolve to `{ns}.{Name}Mock`. Nested models are excluded from the source path but not from the metadata path, so the collision is reachable there. Consider resolving across referenced assemblies rather than through a single-name lookup, and including the containing type in the name.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -48,16 +78,103 @@ public void Execute(GeneratorExecutionContext context)
...
+		if (Generate(context, described) is not { } generated)
+		{
+			return;
+		}
+
+		emitted.Add(key);
+
+		// A generic model's display string carries characters Roslyn rejects in a hint name, and it
+		// rejects them by throwing — which would drop every remaining mock, not just this one.
+		var fileName = InvalidHintNameCharacters.Replace(key.Replace('.', '_'), "_") + ".Mock.g.cs";
```
dr1rrb: ### ➡️ Medium
The hint name is not injective, and a collision takes down every mock rather than one. `key.Replace('.', '_')` followed by the character replace maps `A.B` plus a type `C` and a global type named `A_B_C` to the same file name, and `Ns.Model<T>` collides with a type literally named `Ns.Model_T_`. The `emitted` guard above is keyed on the display string, not on the file name, so both reach `AddSource`, which throws `ArgumentException` ("the hintName of the added source file must be unique within a generator") and surfaces as CS8785, precisely the failure mode the comment two lines up says it is avoiding. On the metadata path the models come from referenced assemblies, so a third-party type name can break the consumer's build.

Worth noting the repository already has the helper for half of this: `PathHelper.SanitizeFileName` lives in `Uno.Extensions.Core.Generators/Common`, which this project links at `Uno.HotTesting.Reactive.Generator.csproj:26`, and six sibling generators call it. It does not give uniqueness on its own, so the fix is likely that helper plus a short hash or ordinal suffix when a sanitized name repeats.

dr1rrb (commented):
➡️ **Medium** — 10 medium, 6 low, no blocker.

The two defects this targets are fixed at the root, and the design call behind it is the right one: giving the mocking generator a second way in, rather than moving emission into the MVUX generator, keeps mock code generation out of Core. The stateless shape asked for previously is genuinely restored, the opt-out cuts both paths at a single point before any enumeration, and the source path's view-model naming does match what the MVUX generator derives. Nothing here blocks a merge.

What the medium findings share is one theme: the change states an anti-drift invariant, then applies it to one of the four rules the two generators must agree on, and it does so by putting the shared file in the core generator under a folder named after the consumer that globs it. Model discovery, view-model naming and constructor selection stay re-implemented, and they have already diverged. The rest are containment issues on the new path: a hint name that is not injective and takes every mock down with it on a collision, two swallowed regex failures whose justifying comment does not hold, and an accessibility fix that only covers the source path.

Two points with no line to attach them to:

- **The verification count does not match the tree.** The description reports "the `Mocking` suite of `Uno.Extensions.Reactive.Tests` 19/19"; at this head the suite holds 15 `[TestMethod]` and no `[DataRow]` (7 + 4 + 4 across `Given_GeneratedMock`, `Given_MockingRuntime`, `Given_SingleProjectMock`). It is the line a reviewer leans on to judge coverage, so it is worth correcting.
- **`specs/013-mvux-mocking-previews/implementation.md` was left behind.** Three of the four spec documents are updated; this one still titles its section 4.2 "Generator (runs in the consumer/test project, metadata-driven)" while the generator gains a source path, and its `rules.md` inventory still stops at `MOCK0001` while `MOCK0002` is added.

Suggested order for a single pass:

1. Share the model-discovery predicate and the view-model naming, and resolve the three attributes by metadata name.
2. Make the hint name unique, set the accessibility on the metadata path, and report a diagnostic instead of swallowing a rejected pattern.
3. Replace the polling wait in the new test, then the doc and spec corrections above.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -83,12 +200,128 @@ IEnumerable<INamedTypeSymbol> Walk(INamespaceOrTypeSymbol ns)
...
+			return Regex.IsMatch(typeName, pattern, RegexOptions.CultureInvariant, RegexMatchTimeout);
+		}
+		catch (ArgumentException)
+		{
+			return false; // Not a valid expression — the MVUX generator reports on the attribute itself.
+		}
+		catch (RegexMatchTimeoutException)
+		{
+			return false;
+		}
```
dr1rrb: ### ➡️ Medium
Both catches return `false` with no signal, and the comment justifying the first one does not hold: `ViewModelGenTool_2.cs:48` and `ViewModelGenTool_3.cs:77` call `Regex.IsMatch` with no `try`, no timeout and no diagnostic, so nothing reports on the attribute. A malformed `[ImplicitBindables("(Model$")]` therefore produces a CS8785 raised by a different component, naming neither the attribute nor the pattern, plus zero mocks here, with nothing connecting the two symptoms.

The timeout has the same shape: one second bounds a single match, not the aggregate. It applies per pattern, per type, on every compilation, so on every keystroke in the IDE, and its expiry is swallowed too, which makes a pathological pattern look like a build that is merely slow and has no mocks. Reporting a diagnostic on the attribute location naming the offending pattern, once per compilation, would cost little and matches how this generator now handles its other bail-outs.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -97,16 +330,44 @@ private sealed class FeedMember
...
 
-		var inputs = new List<FeedMember>();   // OnParameter set
-		var derived = new List<FeedMember>();  // OnFeed set
+		var described = new ModelMock
+		{
+			Model = model,
+			ViewModelName = vm.Name,
+			ViewModelFullName = vm.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat),
+			Constructor = PickConstructor(vm.Constructors),
+		};
```
dr1rrb, line 399: ### ➡️ Medium
`Accessibility` is never set on this path, so it keeps its `"public"` default while the source path derives it from the model. An `internal` model in a referenced assembly is enumerated here, since Roslyn exposes internal types of metadata references, and yields a `public` mock whose `Create` returns an inaccessible view-model: CS0122, or CS0050/CS0051 under `InternalsVisibleTo`. The field's own comment states the invariant without qualifying it to one path, and the only accessibility test added goes through the source path. Setting it from the resolved view-model's `DeclaredAccessibility` here, plus a generator-driver case for an internal model reached through metadata, would close it.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -127,105 +388,189 @@ private sealed class FeedMember
...
+		{
+			// The original defect this generator had was producing nothing without saying so. A model with
+			// feeds but no mockable one is a misclassification often enough to be worth reporting; Info so a
+			// legitimately all-independent model cannot fail a warnings-as-errors build.
+			context.ReportDiagnostic(Diagnostic.Create(
+				NoMockableInput,
+				model.Locations.FirstOrDefault(location => location.IsInSource) ?? Location.None,
+				model.Name));
+			return null;
+		}
```
dr1rrb, line 559: ### 👁 Low
Leaving `emitted` unrecorded until generation succeeds is deliberate and correct: it keeps a model the metadata path could not describe available to the source path. The side effect is that a model reached by both paths and rejected by both is diagnosed twice, so the same MOCK0002 (or MOCK0001) appears against one model in one build. Tracking the models already diagnosed alongside `emitted` keeps the report one per model.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
```diff
@@ -127,105 +388,189 @@ private sealed class FeedMember
...
+		=> constructors
+			.Where(c => !c.IsStatic && IsCandidate(c, includeInternal) && ReadReactiveBindable(c) is not false)
 			.OrderBy(c => c.Parameters.Length)
 			.ThenBy(ParameterTypes, StringComparer.Ordinal)
 			.FirstOrDefault();
-		if (ctor is null)
+
+	private static bool IsCandidate(IMethodSymbol ctor, bool includeInternal)
+		=> ctor.DeclaredAccessibility == Accessibility.Public
+			|| (includeInternal && ctor.DeclaredAccessibility == Accessibility.Internal);
```
dr1rrb: ### 👁 Low
`IsCandidate` accepts only `Public` and `Internal`, but the MVUX generator mirrors `ProtectedOrInternal` constructors as well (`ViewModelGenTool_3.cs:123-125` excludes only private and protected). A model whose only non-private constructor is `protected internal` is therefore reported as having no public constructor for `Create` to null-inject, even though the mirrored constructor would be callable from the same assembly. Including `ProtectedOrInternal` when `includeInternal` is set lines the two up.

src/Uno.Extensions.Reactive.Tests/Mocking/Given_SingleProjectMock.cs
```diff
@@ -0,0 +1,98 @@
...
+		{
+			if (result.Count > 0 && result.Last().Current.Data.IsSome(out var value))
+			{
+				return (IImmutableList<T>)value!;
+			}
+
+			await Task.Delay(20);
+		}
+
+		throw new AssertFailedException("The feed produced no value: the mocked input was never observed.");
```
dr1rrb: ### ➡️ Medium
This wait is a fixed one-second budget spent in 50 polls of 20 ms, with no cancellation and no `[Timeout]`, and a loaded CI agent turns a correctness assertion into a flake whose message blames the mock rather than the deadline. Two further details make it worth replacing rather than tuning:

1. It is a copy of `CurrentItems` in `Given_GeneratedMock` with the opposite timeout semantics: that one returns `null`, and `When_CreateDefault_Then_InputsAreEmpty` relies on the `null`. Unifying the two later would silently invert this test.
2. The test library already ships `WaitForMessages`, reachable from the very `Record()` this helper creates.

Awaiting the recorder's notification, or taking the test's `CancellationToken` with an explicit `[Timeout]`, removes the magic second and makes the deadline readable.

src/Uno.Extensions.Core.Generators/Common/FeedDependencyAnalysis.cs
```diff
@@ -0,0 +1,366 @@
...
+			return (FeedKind.ServiceDependent, derivedFrom, services);
+		}
+		return (FeedKind.Independent, derivedFrom, services);
+	}
+
+	/// <summary>
+	/// Returns the getter/initializer body syntax nodes of a feed member (property expression body,
+	/// getter body, or field initializer).
+	/// </summary>
+	private SemanticModel GetSemanticModel(SyntaxTree tree)
```
dr1rrb, line 236: ### 👁 Low
This `<summary>` describes "the getter/initializer body syntax nodes of a feed member", but it sits on `GetSemanticModel`, which returns a cached `SemanticModel`. The method it actually describes, `GetMemberBodies`, is immediately below and carries no documentation. Moving the block down one member fixes both halves.

src/Uno.Extensions.Core.Generators/Common/FeedDependencyAnalysis.cs
```diff
@@ -0,0 +1,366 @@
...
+					continue;
+				}
+
+				var semanticModel = GetSemanticModel(node.SyntaxTree);
+				InspectEager(body, semanticModel, ctorParamNames, Mark, enclosingMember: null);
+			}
+		}
+
+		return eager;
+	}
```
dr1rrb, line 307: ### 👁 Low
`enclosingMember` is `null` at the only call site, so `Mark` never adds to its `HashSet` and the `kvp.Value.Count > 0` branch in `ViewModelGenTool_3.Mocking.cs` is unreachable: the emitted `CtorDependency` never carries `Members`. The doc comment above also promises a scan of "a ctor body or an instance field/property initializer", while the loop only visits constructor bodies. The behaviour predates this PR, but moving the code into a shared file is the cheapest moment to drop the unused parameter and correct the comment.

doc/Reference/Reactive/testing.md
```diff
@@ -167,8 +176,8 @@ vm.SetMock(RecipeModelMock.Empty with { Steps = ListFeedMock.Error<Step>(new Tim
 > [!IMPORTANT]
 > Mocking is only active for a view-model built by `Create` (which opens a
 > `MockingService.Enable()` scope around construction). Outside such a scope
-> nothing is wrapped, so a shipping application pays no cost — never reference
-> `Uno.HotTesting.Reactive` from a published app head.
+> nothing is wrapped, so a shipping application pays no cost, including one that
+> carries the reference itself.
```
dr1rrb: ### ➡️ Medium
"Pays no cost, including one that carries the reference itself" is true of the runtime wrapping, since nothing is wrapped outside a `MockingService.Enable()` scope, but it now reads as a general statement and the app head does pay something: it gains a runtime package reference, and the generator emits `public sealed record {Model}Mock` and `public static partial class {Vm}Mock` into the application's own assembly. `Uno.HotTesting.Reactive` is marked neither `DevelopmentDependency` nor `PrivateAssets`, so a published single-project app ships the test seam, and `[EditorBrowsable(Never)]` hides it from completion without bounding it. Either narrow the sentence to the runtime cost and say what is emitted, or make the package a development dependency and gate emission on an MSBuild property.

doc/Reference/Reactive/rules.md
```diff
@@ -38,3 +38,16 @@ emits no `{Model}Mock` / `{Vm}Mock` for that model.
...
 Make one of the model's constructors public to get the mock back, or suppress the warning for a model that is
 not meant to be mocked.
+
+## Mock0002
+
+**No mock is generated for a model none of whose feeds is fed by a constructor parameter.**
+
+A `{Model}Mock` replaces the model's _inputs_ — the feeds reading a service the model takes as a constructor
+parameter — and the feeds derived from them. A model whose feeds are all independent has nothing a mock could
+drive, so the generator emits none and says so rather than leaving you to wonder why nothing appeared.
```
dr1rrb, line 48: ### 👁 Low
This says the generator "says so rather than leaving you to wonder why nothing appeared", but MOCK0002 is declared at `DiagnosticSeverity.Info`, and a message-level diagnostic is not printed by `dotnet build` at default verbosity. Choosing `Info` is right, since a legitimately all-independent model must not fail a warnings-as-errors build, so the fix is on this side: say where the message shows up (IDE error list, or `-v normal`), or how to raise it through `.editorconfig`.

src/Uno.HotTesting.Reactive.Generator/Uno.HotTesting.Reactive.Generator.csproj
```diff
@@ -26,5 +26,8 @@
 		<Compile Include="..\Uno.Extensions.Core.Generators\Common\**\*.cs" Link="Common\%(RecursiveDir)%(Filename)%(Extension)" />
 
 		<Compile Include="..\Uno.Extensions.Reactive\**\*Attribute.cs" Link="Attributes\%(RecursiveDir)%(Filename)%(Extension)" />
+
+		<!-- Shared with the MVUX generator: the two must agree on what a model's inputs are (spec 013). -->
+		<Compile Include="..\Uno.Extensions.Reactive.Generator\Mocking\*.cs" Link="Mocking\%(Filename)%(Extension)" />
```
dr1rrb: ### 👁 Low
This glob is the mechanism carrying the "cannot drift apart" invariant, and it is the only non-recursive one in the `ItemGroup`: the three includes above all use `**`. A file added under a subfolder of `Mocking/` would quietly stop being shared, with no error and no failing test. Using the recursive form with the matching `Link` metadata would make the invariant hold by construction.

src/Uno.Extensions.Reactive.Generator/Mocking/FeedMockingAnalysis.cs
```diff
@@ -0,0 +1,366 @@
...
+/// inputs, which are derived, and which constructor parameters are dereferenced eagerly.
+///
+/// It is shared by the two generators that need the same answers and must not drift apart:
+/// <c>ViewModelGenTool_3</c> emits it as <c>[FeedDependency]</c>/<c>[CtorDependency]</c>
+/// metadata for consumers that read the app as a compiled reference, and the mocking generator in
+/// <c>Uno.HotTesting.Reactive</c> (which links this file) runs it directly when the models sit in the
+/// compilation being generated — there the attributes are emitted by a sibling generator and a
+/// generator cannot observe another generator's output.
+/// </summary>
+internal sealed class FeedMockingAnalysis
```
dr1rrb: ### ➡️ Medium
The shared analysis is hosted by the core generator and named after the one consumer that reaches into it. `src/Uno.Extensions.Reactive.Generator/Mocking/` does not exist on `main`; it is created here, and the project that owns it is packed into `Uno.Extensions.Reactive` as `analyzers/dotnet/cs`. So the core analyzer now carries a `Mocking` folder whose only reason to exist is that `Uno.HotTesting.Reactive.Generator.csproj` globs it. The boundary the design notes draw — no mock code generation in the core generator — still holds for emission, but the layout now carries the downstream concern.

What the type does is not mocking. `GetFeedMembers`, `GetCtorParameterNames`, `BuildFieldToParamMap`, `ClassifyFeedMember` and `FindEagerCtorParameters` classify a model's feed members and find constructor parameters dereferenced eagerly. That is what `[FeedDependency]` and `[CtorDependency]` are, and they exist for MVUX whether or not anything is ever mocked — the enum here is already named `FeedKind`, not after mocking.

Two consequences worth weighing:

1. A folder name and path in this project are now part of another package's compilation, with no project reference expressing it. Renaming or moving `Mocking/` breaks a project that does not reference this one, at build time, with nothing in this project to catch it.
2. The repository already has a neutral, symmetric place for exactly this: `Uno.Extensions.Core.Generators/Common/**` is globbed by `Uno.Extensions.Reactive.Generator.csproj:29` and by `Uno.HotTesting.Reactive.Generator.csproj:26`. Putting the analysis there, named for what it computes (`FeedDependencyAnalysis`, say), reuses the sharing mechanism that already exists instead of adding a second one that only points one way.

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
rajamatt: Done

src/Uno.HotTesting.Reactive.Generator/FeedsMockGenerator.cs
rajamatt: Done

src/Uno.Extensions.Core.Generators/Common/FeedDependencyAnalysis.cs
rajamatt: Done

src/Uno.Extensions.Core.Generators/Common/FeedDependencyAnalysis.cs
rajamatt: Done

doc/Reference/Reactive/rules.md
rajamatt: Done, fixed in the docs

PR #3185, opened 2026-09-17
feat(msal): cache the ID token so apps can read the user's claims

kazo0:
The MSAL provider only copied the access token into the Uno token cache, so ITokenCache.TokenAsync(IdTokenKey) was always empty while the OIDC provider populates it. A B2C app had no way to read the user's claims (custom attributes, roles, user flow) for in-app authorization.

- store AuthenticationResult.IdToken under TokenCacheExtensions.IdTokenKey
- Given_MsalAuthentication.When_Login_Then_IdTokenCached (red before, green after)
- assert the ID token survives a silent refresh
- When_TokenResponseHasNoAccessToken_Then_NotAuthenticated pins the B2C openid-only case
- token-leak guard now also asserts the ID token never reaches the logs
- docs: B2C authority/scope/broker/redirect notes and a "Reading the user's claims" section


Kunal22shah (approved):

kazo0:
@Mergifyio backport servicing/7.4

